PR #857, opened 2014-12-23
Add popular links / previous and next links for articles

neoeno:
_Amalgamation of this and #845_

This PR is to introduce the Popular Links part of the Related Content work, which is a panel on the right hand side of articles.

Changes in this PR will probably want a quick review by frontend devs (cc: @benbarnett/@alexwllms ) as we've modified the main layout for all articles.

Work related to bringing in the popular links is behind a feature flag: 'popular_links'. The related for popular links can be found in the following PR https://github.com/moneyadviceservice/cms/pull/116

---

Looks like this:

screenshot 2014-12-23 15 36 34

Code looks roughly as you'd expect, adding methods to `Core::Article` called `next_link`/`previous_link`.

See moneyadviceservice/cms#120


neoeno:
Ready for merge I think.

PR #864, opened 2015-01-07
mount Retirement Income Options engine

asmega:
- Unfortunately I can't use the gem version and have to add the git version. As there is already a gem named `rio` on rubygems. We are not on a recent enough version of bundler where the source on a gem can be specified
- The `category_id` has yet to be decided and is subject to change


benlovell:
:+1:

rnalexander:
:+1:

PR #865, opened 2015-01-07
Remove invalid/unused RequireJS paths

munckymagik:
This is a small change made to look bigger by the second commit that alpha-sorts the sections. See 35885b7 for the meaningful change.

The path alias for the entry-point script for an engine is constructed
using using the value of the `data-engine` attribute from the engine's
layout with "Config" added on the end (see bottom of application.js).

So these two paths for car_cost_tool and pensions_calculator we never
being used and in fact referred to non-existent files.

See: https://trello.com/c/PlIsKwOm


benbarnett:
Cleaner :+1:

munckymagik:
Thanks Ben.

PR #866, opened 2015-01-08
Unconstrained engine layout

asmega:
- This allows engines to be not width constrained and use the full width of the page.
- Rio will use this

### DIFF
--- a/app/controllers/rio_controller.rb
+++ b/app/controllers/rio_controller.rb
@@ -1,4 +1,13 @@
 class RioController < EmbeddedToolsController
+  def parent_template
+    if syndicated_tool_request?
+      response.headers['X-Frame-Options'] = 'ALLOWALL'
+      'layouts/engine_syndicated'
+    else
+      'layouts/engine_unconstrained'
+    end
+  end
+
   protected
 
   def category_id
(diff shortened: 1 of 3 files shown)


app/views/layouts/engine.html.erb
```diff
@@ -12,11 +12,7 @@
 <% end %>
 
 <%= render layout: 'layouts/unconstrained' do %>
-  <div class="l-constrained">
```
benbarnett: Is this intentional? I think this might affect a few other tools

benlovell:
:+1:

app/views/layouts/engine.html.erb
benbarnett: For example https://www.moneyadviceservice.org.uk/en/tools/mortgage-calculator relies on `l-container-tool` being there

asmega: whoops may have checked in the file unintentionally...

asmega: accidental check in has now been removed

app/controllers/rio_controller.rb
```diff
@@ -1,4 +1,13 @@
 class RioController < EmbeddedToolsController
+  def parent_template
+    if syndicated_tool_request?
+      response.headers['X-Frame-Options'] = 'ALLOWALL'
+      'layouts/engine_syndicated'
+    else
+      'layouts/engine_unconstrained'
+    end
+  end
+
```
asmega: I'm not sure it would have much benefit in this context.

PR #869, opened 2015-01-13
serve videos from the frontend

dougdroper:
Move the video's from the public website.
Fetch video content from the CMS and fallback to the public website .

### DIFF
--- a/app/controllers/videos_controller.rb
+++ b/app/controllers/videos_controller.rb
@@ -20,8 +20,10 @@ class VideosController < ApplicationController
   end
 
   def set_related_content
-    @related_content = CategoriesWithRestrictedContents.build(@video.categories,
-                                                              RelatedContent.build(@video))
+    @related_content = CategoriesWithRestrictedContents.build(
+      @video.categories,
+      RelatedContent.build(@video)
+    )
   end
 
   def set_categories
--- a/app/controllers/videos_controller.rb
+++ b/app/controllers/videos_controller.rb
@@ -1,0 +1,33 @@
+class VideosController < ApplicationController
+  decorates_assigned :video, with: ContentItemDecorator
+  decorates_assigned :related_content, with: CategoryDecorator
+
+  include Navigation
+
+  def show
+    @video = interactor.call do
+      not_found
+    end
+
+    set_related_content
+    set_categories
+  end
+
+  private
+
+  def interactor
+    Core::VideoReader.new(params[:id])
+  end
+
+  def set_related_content
+    @related_content = CategoriesWithRestrictedContents.build(@video.categories,
+                                                              RelatedContent.build(@video))
+  end
+
+  def set_categories
+    @video.categories.each do |category|
+      active_category category.id
+      active_category category.parent_id if category.child?
+    end
+  end
+end
--- a/lib/core/interactor/video_reader.rb
+++ b/lib/core/interactor/video_reader.rb
@@ -1,15 +1,13 @@
 module Core
   class VideoReader
-    attr_accessor :id
-
-    private :id=
+    attr_reader :id
 
     def initialize(id)
-      self.id = id
+      @id = id
     end
 
     def call
-      data    = repository.find(id)
+      data  = repository.find(id)
       video = Video.new(id, data)
 
       if video.valid?
--- a/lib/core/interactor/video_reader.rb
+++ b/lib/core/interactor/video_reader.rb
@@ -1,0 +1,36 @@
+module Core
+  class VideoReader
+    attr_accessor :id
+
+    private :id=
+
+    def initialize(id)
+      self.id = id
+    end
+
+    def call
+      data    = repository.find(id)
+      video = Video.new(id, data)
+
+      if video.valid?
+        video.tap do |v|
+          v.categories = build_categories(v.categories)
+        end
+      elsif block_given?
+        yield
+      end
+    end
+
+    private
+
+    def repository
+      Registry::Repository[:video]
+    end
+
+    def build_categories(category_ids)
+      category_ids.map do |category_id|
+        CategoryReader.new(category_id).call
+      end
+    end
+  end
+end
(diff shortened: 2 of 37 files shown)


app/controllers/videos_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  private
+
+  def interactor
+    Core::VideoReader.new(params[:id])
+  end
+
+  def set_related_content
+    @related_content = CategoriesWithRestrictedContents.build(@video.categories,
```
benlovell: Stylistically it might be better to break each argument to `build` on a new line. i.e.

``` ruby
BlehThing.build(
  @video.categories,
  @something.things
)    
```

lib/core/interactor/video_reader.rb
```diff
@@ -0,0 +1,36 @@
+module Core
+  class VideoReader
+    attr_accessor :id
+
+    private :id=
```
benlovell: Use `attr_reader` and in your initializer make the assignment via `@id = id` instead.

lib/core/interactor/video_reader.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def initialize(id)
+      self.id = id
+    end
+
+    def call
+      data    = repository.find(id)
```
benlovell: Alignment of the tokens is a little off compared to the next line.

lib/core/repository/videos/public_website.rb
```diff
@@ -0,0 +1,25 @@
+module Core::Repository
+  module Videos
+    class PublicWebsite < Core::Repository::Base
+      def initialize
+        self.connection = Core::Registry::Connection[:public_website]
+      end
+
+      def find(id)
+        response = connection.get('/%{locale}/videos/%{id}.json' %
```
benlovell: Same as I mentioned before, I think stylistically it's easier to parse each argument on a separate line.

lib/core/repository/videos/public_website.rb
```diff
@@ -0,0 +1,25 @@
...
+                                    { locale: I18n.locale, id: id })
+
+        response.body
+
+      rescue Core::Connection::Http::ResourceNotFound
+        nil
```
benlovell: Is this worth instrumenting anywhere?

spec/controllers/videos_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe VideosController, type: :controller do
+  describe 'GET show' do
+    context 'when a video does exist' do
+      let(:categories) { [] }
+      let(:video) { Core::Video.new('test', title: 'test', body: 'body', categories: categories) }
+
+      before do
+        expect(Core::VideoReader).to receive(:new) do
```
benlovell: This should be `allow`, the difference is subtle but meaningful in terms of what we are trying to express in the spec. The `before` being for setup, rather than specifying expectations on behaviour.

spec/controllers/videos_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+
+      it 'is successful' do
+        get :show, id: video.id, locale: I18n.locale
+        expect(response).to be_ok
+      end
+    end
+
+    context 'when an video does not exist' do
+      before { allow_any_instance_of(Core::VideoReader).to receive(:call).and_yield }
```
benlovell: I tend to double take when I see `allow_any_instance_of` it's usually considered a smell. Can we avoid it here?

spec/lib/core/interactor/video_reader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        it 'instantiates the video with the id and the attributes from the repository' do
+          expect(Video)
+            .to receive(:new).with(id, data).and_call_original
+
+          subject.call
+        end
+
+        context 'when the Video entity is valid' do
+          before do
+            expect_any_instance_of(Video).to receive(:valid?) { true }
```
benlovell: This should be `allow_` and also if we can avoid stubbing any instance that would always be preferable.

spec/lib/core/interactor/video_reader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          end
+
+          it 'returns an Video' do
+            expect(subject.call).to be_a(Video)
```
benlovell: This is probably a lay over from your TDDing this feature but asserting on the type is no longer necessary now the structure/interface of this thing is well covered elsewhere.

spec/lib/core/interactor/video_reader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          end
+        end
+
+        context 'when the Video is invalid' do
+          before do
+            expect_any_instance_of(Video).to receive(:valid?) { false }
```
benlovell: Same here with the allow/expect issue and also the `any_instance_of`.

spec/lib/core/interactor/video_reader_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          it_has_behavior 'optional failure block'
+        end
+
+        context 'when there is a category' do
+          let(:category) { 'foo' }
+          let(:categories) { [category] }
+          let(:category_entity) { double }
+
+          before do
+            allow_any_instance_of(CategoryReader).to receive(:call) { category_entity }
```
benlovell: Cool that this is `allow_` but even cooler to drop the `any_instance_of`

benlovell:
Solid :heart_eyes: 

Just a few stylistic things I raised.

darokel:
Just a minor stylistic comment I made but all gooood :shipit:

benlovell:
:+1: :ship: :shipit:

tomas-stefano:
:+1: :+1:

PR #870, opened 2015-01-13
adding related links

JiggyPete:
This PR is to provide a set of related links to an article coming from contento.

The functionality is hidden behind the feature flag
related_links: true


spec/lib/core/entity/article_spec.rb
```diff
@@ -224,6 +224,69 @@ module Core
...
+              'path' => '/next-related-link'
+            }
+          ]
+        }
+      end
+
+      context 'provide data' do
+        it 'has 2 article links' do
+          expect(subject.related_links.length).to eq(2)
+          expect(subject.related_links.first).to be_an_instance_of(ArticleLink)
```
benlovell: This and the following assertions on type are no longer necessary now the interface of this is specified elsewhere, I guess these are a lay over from your TDDing?

benlovell:
:+1: other than that minor comment.

darokel:
:ship:

PR #875, opened 2015-01-16
Add blog links

darokel:
Adds links to the blog in the category section of the home page and primary footer. This is hard coded for now as the blog is going live on Monday and there is no easy way (or know one knows how) to add a new category in the Category Manager Application. Since that category data is being migrated to the CMS in the next week, this will also be added when that is done so it's dynamic like the others.

### DIFF
--- a/app/views/shared/_directory.html.erb
+++ b/app/views/shared/_directory.html.erb
@@ -8,4 +8,11 @@
     <p class="directory__description"><%= item.description %></p>
   </div>
   <% end %>
+  <div class="directory__item">
+    <%= heading_tag(level: heading_level ||= 3, class: 'directory__title') do %>
+      <%= link_to t("home.show.blog_title"), t("footer.blog_link"), class: 'directory__link' %>
+    <% end %>
+
+    <p class="directory__description"><%= t("home.show.blog_description") %></p>
+  </div>
 </div>
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -243,7 +243,7 @@ cy:
     partners: Partneriaid
     partners_link: /cy/categories/Partners
     blog: Eich blog Cyngor Ariannol
-    blog_link: https://blog.moneyadviceservice.org
+    blog_link: https://blog.moneyadviceservice.org.uk
     tools_and_resources: Offer a chyfrifianellau
     tools_and_resources_link: /cy/categories/tools-and-calculators
     site_map: Map safle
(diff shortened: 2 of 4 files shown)


app/views/shared/_directory.html.erb
```diff
@@ -8,4 +8,11 @@
     <p class="directory__description"><%= item.description %></p>
   </div>
   <% end %>
+  <div class="directory__item">
+    <%= heading_tag(level: heading_level ||= 3, class: 'directory__title') do %>
+      <%= link_to t("home.show.blog_title"), 'https://blog.moneyadviceservice.org', class: 'directory__link' %>
```
neoeno: I know this code will probably have a short lifetime, but is there a reason why this uses a plain old URL here and going to the locales [here](https://github.com/moneyadviceservice/frontend/pull/875/files#diff-9c9d02deb56445d7f85f9bc4c16c9a0eR38)?

(Since it's not a localized URL, I'd tend to just use a straight URL everywhere.)

neoeno:
Other than that :+1:

rnalexander:
:+1:

config/locales/en.yml
```diff
@@ -239,8 +241,8 @@ en:
     media_centre_link: /en/static/media-centre
     partners: Partners
     partners_link: /en/categories/partners
-    news: News & features
-    news_link: /en/news
+    blog: Your Money Advice blog
+    blog_link: https://blog.moneyadviceservice.org
```
asmega: is this link right?

app/views/shared/_directory.html.erb
```diff
@@ -8,4 +8,11 @@
     <p class="directory__description"><%= item.description %></p>
   </div>
   <% end %>
+  <div class="directory__item">
+    <%= heading_tag(level: heading_level ||= 3, class: 'directory__title') do %>
```
jongilbraith: Should that be ``heading_level || 3``?

config/locales/cy.yml
```diff
@@ -240,8 +242,8 @@ cy:
     media_centre_link: /cy/static/canolfan-gyfryngau
     partners: Partneriaid
     partners_link: /cy/categories/Partners
-    news: Newyddion ac erthyglau
-    news_link: /cy/news
+    blog: Eich blog Cyngor Ariannol
+    blog_link: https://blog.moneyadviceservice.org
```
asmega: needs .uk on the end?

config/locales/en.yml
darokel: Crap yeah thanks for spotting that

app/views/shared/_directory.html.erb
darokel: I've copied the exact same markup used to produce the other links so they're all styled the same.

PR #876, opened 2015-01-19
add syndicated layout

asmega:
The requirement is when budget planner (or other tool) is syndicated (in an iframe) the user needs to be able to  authenticate. For example they want to save their data which then sends the user to a login screen. The old behaviour is shown in the screenshot below which uses public website and the old budget planner.

old-bp-login

The light box/pop up dialog has been removed and is no longer available. I'm also not aware of such a dough component existing to replace it. Instead we send the user straight to the login page and then redirect them back to where they came from. See below..

sign-in-with

However we do not want to render the entire site with header and footer when we are in the iframe we want to render the minimal amount in the iframe. The pull request will make the login page look like below...

sign-in-without

Changes will need to be made to `scripts` repo to expose `/en|cy/users` when navigating this url from the `partner-tools` subdomain.

Is there a nicer way to implement this?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,8 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   add_flash_types :success
 
+  layout :check_syndicated_layout
+
   include Authentication
   include Chat
   include Localisation
@@ -70,6 +72,14 @@ class ApplicationController < ActionController::Base
   end
 
   private
+
+  def check_syndicated_layout
+    if syndicated_tool_request?
+      'syndicated'
+    else
+      'application'
+    end
+  end
 
   def category_tree
     @category_tree ||= Core::CategoryTreeReader.new.call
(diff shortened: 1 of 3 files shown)


darokel:
A modal would be nice as it would make the user feel they're still within the context of their original source.

asmega:
@darokel Do we have a dough component for that?

tomas-stefano:
@asmega I think @ourmaninamsterdam or other frontend dev could answer that.

ourmaninamsterdam:
@asmega We have a [dough Dialog component](https://github.com/moneyadviceservice/cms/blob/master/app/assets/javascripts/comfortable_mexican_sofa/admin/modules/Dialog.js) in the CMS that is under contention to be moved across to dough.

benbarnett:
We need to consider the usability on mobile though - as that's not a requirement for the CMS right @ourmaninamsterdam?

ourmaninamsterdam:
@benbarnett Precisely. Accessibility too, as it will definitely need some work before it can be used in a customer-facing environment.

asmega:
@darokel So it looks like a modal is out of the mix. Regarding the context typically syndicated tools have the mas logo at the top of the tool.

rnalexander:
If modal's not ready to go I say go ahead with the non-modal solution.  :+1:

benlovell:
LGTM :shipit:

PR #880, opened 2015-01-21
Make failing cucumber fail CI

neoeno:
Call cucumber directly to make it return a non-zero exit status.

Previously, running `test.sh` would run `rake cucumber`, which always returns a zero exit code (for some reason). This meant test failures wouldn't fail CI.

### DIFF
--- a/config/initializers/disable_minitest_hook.rb
+++ b/config/initializers/disable_minitest_hook.rb
@@ -1,0 +1,16 @@
+# Minitest hijacks the exit code set by cucumber.
+# Minitest.autorun sets up a series of at_exit hooks that
+# interfere with it.
+
+# In certain circumstances, this will make `rake cucumber`
+# exit with 0 every time, making our build pass erroneously.
+
+# Since we don't use minitest in any capacity and it seems
+# to only be here to make life interesting - we're going to
+# patch it out.
+
+if defined? Minitest
+  Minitest.module_eval do
+    def self.autorun; end
+  end
+end
--- a/test.sh
+++ b/test.sh
@@ -19,4 +19,4 @@ bowndler install --production --config.interactive=false
 
 RAILS_ENV=development rake karma:install karma:run_once
 rake spec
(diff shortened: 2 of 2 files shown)


neoeno:
NB: merging this will trigger a test failure on CI, because there are currently failing cukes.

benlovell:
Ugh. Good catch. :+1:

munckymagik:
If we do this, should we also remove `lib/tasks/cucumber.rake`? Do we know why it wasn't doing the right thing?

benlovell:
Something somewhere in that chain might be handling `at_exit` and clobbering the real exit code. Maybe.

asmega:
Merging #883 first should make the build pass

neoeno:
#883 is merged, which means all the features should pass now.

However, I'm currently working on a different track on this, having observed some weird behaviour. Will check in here afterwards.

neoeno:
OK, so I spent an inordinately long time working out what the root of this problem is. I think I have about 3/4 of it in my head now.

But basically — cucumber depends on minitest, which sets an `at_exit` hook, which interferes with Cucumber's exit code. The issue is only triggered by some line in the guts of DelayedJob (which is used by email_spec, which we use in `password_reset.feature`).

I've not been able to piece together precisely what's happening, but I know those three things are true.

So I wrote an alternative solution that allows us to keep the rake task as it currently works. Since we're not using Minitest, it should be harmless... but you can pick whichever solution you prefer.

benlovell:
Yikes, so basically what I'd pointed out a few days ago... Good investigative work finding the exact root of it though! :+1:

neoeno:
Ha! Fair point. But only some of the time...

Also, I noticed `newsletter_signup.feature` is still halting cucumber sometimes. At least now it should flag that as a failed build...

asmega:
:+1:  nice work Sherlock

munckymagik:
Crap, forgot to squash first. Soz :(

PR #883, opened 2015-01-22
fix cucumber race condition

asmega:


benlovell:
:+1:

neoeno:
I was a little unsure about how this would work, but ran it and now understand `Timeout.timeout` puts a limit on the length of time that the loop will run for.

:+1:

PR #887, opened 2015-01-29
move x frame options to ApplicationController

asmega:
Regardless of this change this presents the vulnerability of
clickjacking when tools are syndicated


esbel:
looks good :+1:

tomas-stefano:
:+1:

PR #889, opened 2015-02-02
adding Newsletter injection script

Guntrisoft:
- uses dough loader to kick off javascript component
- component injects newsletter div into the right
  place within an article and adds a class
  which displays it to the user
- altered where the newsletter div gets included
  conditionally loading it in an article
  template as long as the current URL
  doesn't match an excluded list of slugs
  or is a welsh article


app/controllers/articles_controller.rb
```diff
@@ -35,4 +36,16 @@ def set_categories
       active_category category.parent_id if category.child?
     end
   end
+
+  def set_show_newsletter_signup
+    exclusions = YAML.load(
+      File.read(File.expand_path("#{Rails.root}/config/newsletter_article_exclusion_list.yml", __FILE__))
```
asmega: is this going to read the file from disk on every request?

Guntrisoft: Probably will! - I'm open to suggestions on how to avoid that, (I'm no rails developer :))

Guntrisoft:
This is eventually going to be replaced by some code which will allow admin users to manage this 'exclusions' list - so this is just a solution until that happens.

Guntrisoft:
@benbarnett welcome to comment on this :)

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+  */
+  DoughBaseComponent.extend(Newsletter);
+
+  /**
+  * Set up and populate the model from the form inputs
+  * @param {Promise} initialised
+  */
+  Newsletter.prototype.init = function(initialised) {
+    this._initialisedSuccess(initialised);
+    this.showInArticle($);
```
benbarnett: Can you add `return this;` to make this chainable?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  Newsletter.prototype.showInArticle = function($) {
+    var $panel = this.getPanel($);
+    var $article = $('.editorial main');
```
benbarnett: Should/can this be namespaced at all? Or targetted in a more generic way via a data attribute?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!$panel.length || !$article.length) {
+      setTimeout(function () {
+        this.showInArticle($);
```
benbarnett: `this` will be `window` here and be undefined
also: `function()` remove spaces between function and brackets

benbarnett: Is there any other way we can do this? Quite possible we'd end up in an infinite loop

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    this.moveToArticle($panel, $article, $);
+
+    $panel.addClass('news-signup-test--with-image news-signup-test--in-article');
+  };
+
+  Newsletter.prototype.moveToArticle = function($form, $article, $) {
+    var article = $article[0];
+    var placed = false;
```
benbarnett: Should be comma separated with a single `var` definition

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+    $form.css('clear', 'both');
+
+    $article.find('> h2, > h3').each(function (i, el) {
+      if (placed) {
+        return;
+      }
+
+      if (el.offsetTop >= minimumPosition) {
+        $form.insertBefore(el);
+        placed = true;
```
benbarnett: Would this be more readible if you just `return`'d out of the loop here rather than the separate if statement at the top?

app/controllers/articles_controller.rb
asmega: Personally I would create a new class that would return the array of articles and not bother with the yml file. That object can change when you want it to read from another source (i.e. something a admin/non-dev can change)

tomas-stefano: Another thing just for showing an improvement. You can make the same thing with:

```ruby
 File.read(Rails.root.join('config', 'newsletter_article_exclusion_list.yml'))
```

tomas-stefano: Now that I saw the entire code, you can improve a little more:

```ruby
YAML.load_file(Rails.root.join('config', 'newsletter_article_exclusion_list.yml'))
```

app/assets/javascripts/components/Newsletter.js
Guntrisoft: @stevoland - any idea why this bit was in the optimisely script? Not sure on a case example of why the panel or article would ever be missing at this point?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+    Newsletter.baseConstructor.call(this, $el, config, defaultConfig);
+  };
+
+  /**
+  * Inherit from base module, for shared methods and interface
+  */
+  DoughBaseComponent.extend(Newsletter);
+
+  /**
+  * Set up and populate the model from the form inputs
```
ourmaninamsterdam: This comment doesn't seem relevant.

app/assets/javascripts/components/Newsletter.js
ourmaninamsterdam: How come jQuery `$` is being passed in?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    });
+
+    if (!placed) {
+      $form.insertAfter($article.children().last());
+    }
+  };
+
+  Newsletter.prototype.getPanel = function($) {
+    return $('.news-signup-test');
```
ourmaninamsterdam: Maybe this should be specified via a `data-dough-newsletter-panel`.

Guntrisoft:
@asmega @benbarnett @ourmaninamsterdam Welcome comments again, on latest changes!

app/controllers/articles_controller.rb
```diff
@@ -35,4 +36,14 @@ def set_categories
       active_category category.parent_id if category.child?
     end
   end
+
+  def set_show_newsletter_signup
+    exclusions = ::NewsletterExclusions::slugs
+
+    exclusions.reject! do
+      |slug| slug != params[:id]
```
asmega: I think it's the first time I've seen it written like that!

Guntrisoft: Whoops, see what you mean, will alter!

lib/newsletter_exclusion.rb
```diff
@@ -0,0 +1,93 @@
+module NewsletterExclusions
```
asmega: Are you sure this works? Is `NewsletterExclusions` actually being loaded?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,60 @@
...
+  */
+  Newsletter.prototype.init = function(initialised) {
+    this._initialisedSuccess(initialised);
+    this.showInArticle();
+    return this;
+  };
+
+  Newsletter.prototype.showInArticle = function() {
+    var $panel = this.getPanel();
+    var $article = $('[data-article]');
```
ourmaninamsterdam: Should probably be `[data-dough-newsletter-article]` to keep it consistent with other Dough components.

ourmaninamsterdam: Should be separate comma-delimited, `var` declarations.

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,60 @@
...
+  * @param {Promise} initialised
+  */
+  Newsletter.prototype.init = function(initialised) {
+    this._initialisedSuccess(initialised);
+    this.showInArticle();
+    return this;
+  };
+
+  Newsletter.prototype.showInArticle = function() {
+    var $panel = this.getPanel();
```
ourmaninamsterdam: As `getPanel` returns `$('[data-dough-component="Newsletter"]')`, it could just be `$panel = this.$el`.

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,60 @@
...
+
+    $panel.addClass('news-signup-test--with-image news-signup-test--in-article');
+  };
+
+  Newsletter.prototype.moveToArticle = function($panel, $article) {
+    var article = $article[0],
+        placed = false,
+        minimumPosition = article.offsetHeight / 4;
+
+    $panel.css('clear', 'both');
```
ourmaninamsterdam: ... `$panel` could then be replaced by `this.$el`.

ourmaninamsterdam:
Will there be any JS tests?

app/assets/javascripts/components/Newsletter.js
```diff
@@ -0,0 +1,60 @@
...
+
+    this.moveToArticle($panel, $article);
+
+    $panel.addClass('news-signup-test--with-image news-signup-test--in-article');
+  };
+
+  Newsletter.prototype.moveToArticle = function($panel, $article) {
+    var article = $article[0],
+        placed = false,
+        minimumPosition = article.offsetHeight / 4;
```
ourmaninamsterdam: The `article` variable is only used once so maybe `$article[0]` would be better.

lib/newsletter_exclusion.rb
Guntrisoft: Seems to be fine as the lib folder is in load path...

app/views/articles/show.html.erb
```diff
@@ -13,8 +13,13 @@
 
 <div class="l-article-3col-main editorial">
   <div class="l-main__row">
-    <main class="l-main__cell-full" role="main">
+    <main class="l-main__cell-full" role="main" data-article="true">
```
ourmaninamsterdam: This could be `[data-dough-newsletter-article]` so the association with the Newsletter component is clearer.

lib/newsletter_exclusion.rb
asmega: :+1:

tomas-stefano: The lib folder is in load_path and but the file is named newsletter_exclusion and the module NewsletterExclusions (with the 's'). Are you sure that is working? But working or not I would rename the file to newsletter_exclusions.rb (with 's'). :)

lib/newsletter_exclusion.rb
```diff
@@ -0,0 +1,93 @@
...
+      'using-an-equity-release-scheme-to-fund-your-care',
+      'paying-your-own-care-costs-but-the-moneys-run-out',
+      'downsizing-your-home-to-fund-your-long-term-care',
+      'if-your-baby-is-stillborn',
+      'if-the-person-you-want-to-help-has-lost-mental-capacity',
+      'putting-your-affairs-in-order',
+      'preparing-for-illness-old-age-and-death',
+      'choosing-your-executor'
+    ]
+  end
```
tomas-stefano: I would move this to some data store instead hard-coded data on a module that exclude slugs from the newsletter.

lib/newsletter_exclusion.rb
tomas-stefano: Since is not including anything on any class I would rename this to a class rather than a module.

lib/newsletter_exclusion.rb
Guntrisoft: This is eventually (at some point later this week I believe!) going to be added as a flag against an article so this can be managed in the db/admin rather than as a hard coded list.

tomas-stefano: Ok, so if is temporarily until the flag :+1:

Guntrisoft:
@ourmaninamsterdam added some JS tests
@tomas-stefano renamed that class (even though it was working ok, you're right it should be same name as class)

tomas-stefano:
@Guntrisoft :+1:

ourmaninamsterdam:
:+1:

PR #894, opened 2015-02-05
[DO NOT MERGE] Get categories from contento

dougdroper:


lib/core/repository/categories/cms.rb
```diff
@@ -0,0 +1,29 @@
...
+      rescue
+        fallback_repository.find(id)
+      end
+
+      private
+
+      attr_accessor :connection, :fallback_repository
+    end
+  end
+end
```
tomas-stefano: I wonder if we can organize this in a superclass. I start here with action plain and I already thinking on the next time will be exactly same things with a different 'resource'.

dougdroper: Yes could do, I wonder if it is worth only doing when we actually have another 'resource' though.

tomas-stefano: Ow, yes. Here on this PR is fine :+1: I was just using this opportunity to start the discussion and heard your opinion :)

tomas-stefano:
:+1:

lib/core/repository/categories/cms.rb
dougdroper: yes cool, sounds good :+1:

lib/core/repository/categories/cms.rb
```diff
@@ -0,0 +1,29 @@
+module Core::Repository
+  module Categories
+    class CMS < Core::Repository::Base
+      def initialize(options = {})
+        self.connection = Core::Registry::Connection[:cms]
+        self.fallback_repository = options[:fallback]
+      end
+
+      def all
+        response = connection.get('/%{locale}/categories.json' % { locale: I18n.locale })
```
benlovell: URI template (RFC 6570) would be nice anywhere you perform these kind of expansions. There are loads of different gems that do template expansion. I've used `uri_template` before.

app/views/categories/_child_categories.html.erb
```diff
@@ -1,4 +1,4 @@
-<% contents.each_with_index do |child_category, i| %>
+<% contents.reject{|c| c.object.type == 'guide' }.each_with_index do |child_category, i| %>
```
benlovell: A couple things. You have this retrieval logic that has leaked out into the view. If it's meaningful it should live in the domain. Secondly - I wonder if `c.guide?` is a thing that encapsulates this check, and if not, could it be? It would let you golf the down to `reject(&:guide?)`

dougdroper: yes cool, agreed. Will rework it.

lib/core/repository/categories/cms.rb
dougdroper: yeah cool, there are a few places that do this kind of expansion, so might be worth changing them all in one fell swoop in a separate PR.

baphled:
:+1:

lib/core/repository/categories/cms.rb
benlovell: :+1:

benlovell:
:ship:

dougdroper:
excellent, running migration on comfy production, then I will merge this in.

lib/core/repository/categories/cms.rb
tomas-stefano: I made the structure on the PR #896 Subclassing from CMS this everything will work :). But I will wait for this PR to be merged on master and change this in my branch, ok?

PR #895, opened 2015-02-06
Remove CI rspec config to the main file and move to CI test file

tomas-stefano:
Each time that you run the specs, they run with the build config and you need to overwrite. So let's move this config to the CI test script and add to .rspec only the default config that everybody uses.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,9 +1,2 @@
 --color
 --require spec_helper
---format html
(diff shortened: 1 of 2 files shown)


misaka:
:+1:

dougdroper:
:+1:  I do prefer that format

PR #900, opened 2015-02-11
Update RIO

stevoland:


asmega:
:+1:

Gemfile
```diff
@@ -50,7 +50,7 @@ gem 'decision_trees', '~> 2.0.0'
 gem 'feedback', '~> 0.3.0'
 gem 'mortgage_calculator', '~> 1.3.5'
 gem 'pensions_calculator', '~> 0.2'
-gem 'rio', git: 'git@github.com:moneyadviceservice/rio.git', ref: 'bb267ca11932fd7958077d89f9802c9ae2d60d5b'
+gem 'rio', git: 'git@github.com:moneyadviceservice/rio.git', ref: '35fc40dd3adbd31124661ea66278fdad3bd6c23b'
```
esbel: please commit updated `Gemfile.lock`, as it still points to `bb267ca11932fd7958077d89f9802c9ae2d60d5b` revision

stevoland: @esbel Thanks, I've force pushed an update. I find this Ruby config business odd. Seems like there's 2 sources of truth for config?

esbel: Thanks.

The purpose of Gemfile and Gemfile.lock is a bit different. Generally speaking, in the Gemfile you should state a minimal set of requirements that should be met by your libraries. On the other hand, Gemfile.lock is meant to specify working configuration with all dependencies resolved (and usually is autogenerated).

I think that in this case, it might be a good idea to remove `ref` option from `Gemfile` and use `bundle update` when we want to update rio gem version.

stevoland: @esbel thanks - makes sense. @asmega what do you think about removing the `ref` from `Gemfile` ^ ?

stevoland:
@esbel Ok, I've removed `ref` from `Gemfile`, cheers

esbel:
thanks @stevoland!

PR #902, opened 2015-02-11
Enable welsh translation links on pages that comes from CMS

tomas-stefano:
The welsh articles today is not working for all pages that comes from CMS. So let's enable this!! :)


asmega:
:+1:

dougdroper:
:+1:

PR #910, opened 2015-02-13
Sort companies alphabetically

benbarnett:


darokel:
:+1:

app/helpers/debt_management_helper.rb
```diff
@@ -0,0 +1,5 @@
+module DebtManagementHelper
+  def companies_list
+    t('companies.companies').sort { |a, b| a <=> b }
```
esbel: `.sort { |a, b| a <=> b }` is equal to just `.sort`

benbarnett: Ah right cool, I'll remove then

benbarnett: Removed

darokel: Good call

esbel:
thanks :+1:

PR #917, opened 2015-02-18
force removal of bower_components directory

asmega:
-f used to suppress failure even if the directory does not exist


esbel:
:+1:

benbarnett:
:+1:

PR #920, opened 2015-02-18
Pin pensions calculator.

neoeno:
We're doing some heavy lifting here, big removals, large change.

This is just to make sure no one runs `bundle update` and pushes it all into production.


benlovell:
:+1:

benbarnett:
:+1:

esbel:
why not >= 0.2.3.521?

benbarnett:
@esbel I think this change is here to prevent the pension calculator changes getting accidentally deployed to production, so its pinned exactly to this 'safe' version

neoeno:
That's right!

esbel:
ok @benbarnett, I wasn't sure about that part, thanks

PR #928, opened 2015-02-23
Fixes missing text in TabSelector - Dough issue #214

ourmaninamsterdam:
See #214 for background information.

Same change has been applied on Dough: https://github.com/moneyadviceservice/dough/pull/220

@moneyadviceservice/frontend

### DIFF
--- a/app/assets/stylesheets/components/dough_theme/tab_selector/_tab_selector.scss
+++ b/app/assets/stylesheets/components/dough_theme/tab_selector/_tab_selector.scss
@@ -64,6 +64,7 @@
       }
 
       .tab-selector__trigger {
+        @extend %type;
         border-width: 0;
         padding-right: 35px;
         padding-left: 12px;


ourmaninamsterdam:
@Guntrisoft @benbarnett  I've put back the `font-size: 0` (I think you're right in saying removing this could cause more issues) and included the missing `@extend %type;`. 

Dough's `tab_selector.scss` will still have the missing text problem as it doesn't have anyway of defining rem/px font sizes, as this is done with Yeast via the `body()` mixin.  So we either leave Dough as it is, and rely on users adding their own font size in their own 'theme' (this would mean updating tools that are currently relying on `TabSelector`), or we add a `body()`-style mixin so we can define rem/px  font sizes in Dough and use this to set the font size on the `tab-selector__trigger`.

benbarnett:
What do you reckon? Dough could probably use a `body()` style mixin, seems generic enough?

ourmaninamsterdam:
I think it would be a good route as without it it could become arduous adding in font-sizes manually or via Yeast.

benbarnett:
Yeah I think so too, sounds good to me

Guntrisoft:
I think the body mixin is a nice way to go.

At the BBC, the GEL guidlines http://www.bbc.co.uk/gel/web/building-blocks/typography/type-sizes dictated clear font sizes and line-height which were put into placeholders in Sass and each given a name.

Then you just @extend %blah and that particular bit of text then responsively changed font size through the viewport widths.

It then avoided 100 million different body(X,Y) across the code - however at least body() is a step in the right direction.

Appreciate that kind of way of doing things needs huge collaboration with the designers to adhere to a set number of font size/line heights!

ourmaninamsterdam:
I agree, I'm not a massive fan of lots of `body()` includes.

Perhaps we just give the Dough components a basic text formatting using `@extend %type;` (as per Yeast convention) and then leave it to the user to adjust the `font-size` in the 'theme' layer, if they require.

So we have *assets/lib/_typography.scss* with:

```scss
@mixin body($font-size, $line-height) {
  @if $responsive == true {
    font-size: ($font-size / 16) + rem;
    line-height: ($line-height / 16) + rem;
  } @else {
    font-size: $font-size + px;
    line-height: $line-height + px;
  }
}

%type {
  @include body(16, 24);
}
```
```scss
.tab-selector__trigger {
  @extend %type;
  ...
}
```
This would allow Yeast to override `%type` in the frontend.

ourmaninamsterdam:
Have pushed the fix to Dough: https://github.com/moneyadviceservice/dough/pull/220

ourmaninamsterdam:
@Guntrisoft @benbarnett Any thoughts? Have pushed a fix to the related Dough PR.

benbarnett:
Big fan :+1:

ourmaninamsterdam:
@Guntrisoft Could you have a gander at this one when you've got a sec please? It's the sister PR to the Dough fix.

Guntrisoft:
:+1:

PR #931, opened 2015-03-03
Remove all articles from blacklist

tomas-stefano:
* There are some harded coded pages that are blocked from preview.
Please, let's remove that.

### DIFF
--- a/config/initializers/valid_resource.rb
+++ b/config/initializers/valid_resource.rb
@@ -1,11 +1,9 @@
 class ValidResource
   BLACKLIST = {
-    article:     %w(),
     category:    %w(partners
                     partners-uc-banks
                     partners-uc-landlords
-                    resources-for-professionals-working-with-young-people-and-parents),
-    static_page: %w()
+                    resources-for-professionals-working-with-young-people-and-parents)
   }
 
   attr_accessor :type
--- a/config/initializers/valid_resource.rb
+++ b/config/initializers/valid_resource.rb
@@ -1,15 +1,6 @@
 class ValidResource
   BLACKLIST = {
-    article:     %w(about-our-debt-work am-ein-gwaith-dyled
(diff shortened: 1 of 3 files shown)


esbel:
:+1:

dougdroper:
As discussed, we need to be careful that we are not hiding the intention of the page :)

tomas-stefano:
Yes, I send to the business this blacklisted pages. I will wait and see for preview and published states.

tomas-stefano:
@dougdroper we decide to remove all articles from blacklist and test on QA all this pages. Can you review now?

config/routes.rb
```diff
@@ -111,8 +111,7 @@ def not_implemented
               end
 
     get '/:page_type/:id/preview' => 'articles_preview#show',
-        page_type: /articles|action_plans|news|videos|corporate|tools/,
-        constraints: ValidResource.new(:article)
```
dougdroper: do we need to remove the constraints from the article route?

tomas-stefano: There is no articles anymore. The call will make anything.

tomas-stefano: There are move valid resource call. I will remove all blacklisted articles Valid resource that is being called.

tomas-stefano: Removed!!!!!

config/initializers/valid_resource.rb
```diff
@@ -1,20 +1,9 @@
 class ValidResource
   BLACKLIST = {
-    article:     %w(about-our-debt-work am-ein-gwaith-dyled
-                    debt-publications cyhoeddiadau-ar-ddyledion
-                    partners-overview-parhub
```
dougdroper: cool, there are some redirects in scripts, that we may need to double check that they are still needed, but :+1:

dougdroper:
:+1:

PR #933, opened 2015-03-03
Render extra content on corporate articles if partials exists.

tomas-stefano:
* This is the basic structure of the Partners/Terms and Conditions and other
corporate pages.
* Today CMS doesn't support forms and some other type of dynamic context, so this extra content is needed.

### DIFF
--- a/app/decorators/corporate_decorator.rb
+++ b/app/decorators/corporate_decorator.rb
@@ -1,0 +1,26 @@
+class CorporateDecorator < ContentItemDecorator
+  attr_reader :extra_content_path
+  delegate :lookup_context, to: :h
+
+  def initialize(*)
+    super
+
+    @extra_content_path = 'corporate/extra'
+  end
+
+  def extra_content_partial
+    "#{extra_content_path}/#{extra_content_partial_name}"
+  end
+
+  def extra_content?
+    path = Array(extra_content_path)
+
+    lookup_context.template_exists?(extra_content_partial_name, path, true)
+  end
+
+  private
+
+  def extra_content_partial_name
+    slug.underscore
+  end
+end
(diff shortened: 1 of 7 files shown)


app/decorators/corporate_decorator.rb
```diff
@@ -0,0 +1,25 @@
+class CorporateDecorator < ContentItemDecorator
+  attr_reader :extra_content_path
+
+  def initialize(object, options = {})
```
benlovell: You could reduce this signature down to `initialize(*)` and this would be passed along to `super`.

tomas-stefano: Good idea! Will do it.

spec/decorators/corporate_decorator_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  include Draper::ViewHelpers
+
+  subject(:decorator) { described_class.decorate(content_item) }
+
+  let(:content_item) { instance_double(Core::Article, id: 'bob', slug: 'syndicated') }
+
+  describe '#extra_content?' do
+    context 'when article has partial' do
+      before do
+        expect_any_instance_of(ActionView::LookupContext).to receive(:template_exists?)
```
benlovell: This feels cumbersome, can this behaviour not be observed or triggered by other means? Hooking into rails internals is generally a bad idea :tm:

tomas-stefano: I will extract the lookup context to a private method and setup expectations from that perspective.

tomas-stefano: Done

app/decorators/corporate_decorator.rb
tomas-stefano: Done

benlovell:
:+1:

dougdroper:
:+1:

PR #934, opened 2015-03-04
add retirement landing page

asmega:


spec/controllers/retirements_controller.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.describe RetirementsController, type: :controller, features: [:rio] do
+  describe '#index' do
+    it 'works' do
```
esbel: change it something than includes success or 200 :D

esbel:
:+1:

Tr4pSt3R:
Might want to add this to the controller test? 
```
expect(response).to render_template("index")
```
just an addition. All looks good to go. :+1:

asmega:
If you delete the view right now the test will probably fail, so it would be a redundant test

asmega:
Thanks!

Tr4pSt3R:
:+1: :100:

PR #935, opened 2015-03-04
change rio mount point to /retirement-income-opt..

asmega:
/retirement-income-options

### DIFF
--- a/spec/controllers/mount_controller_spec.rb
+++ b/spec/controllers/mount_controller_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe MountController, type: :controller do
 
   describe '#alternate_url' do
     it 'returns an alternate_url if one is available' do
-      expect(subject).to receive(:url_for).with('script_name' => '/cy/alternative_engine')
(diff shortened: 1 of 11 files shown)


spec/controllers/mount_controller_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   describe '#alternate_url' do
     it 'returns an alternate_url if one is available' do
-      expect(subject).to receive(:url_for).with('script_name' => '/cy/alternative_engine')
+      expect(subject).to receive(:url_for).with('script_name' => '/cy/alternative_engine').and_return('/cy/alternative_engine')
```
esbel: Would following do the trick?

`expect(subject).to receive(:url_for).with('script_name' => '/cy/alternative_engine').and_call_original`

asmega: unfortunately not. i've not setup all the params so the route fails.

esbel: ok

esbel:
:+1:

PR #946, opened 2015-03-12
Migrates tools.js and dependencies from public_website

picandocodigo:
As public_website will be decommissioned at some point in the future we are migrating this over to the responsive site. `tools.js` is referenced by the html that partners use to embed a syndicated tool, it is also used by the test_harness repo which allows us to mimick a partner who is embedding a tool.

We found a potential trip wire on the public_website project. This code on `build.sh` and one of the Rake tasks:

```
 deploy.rake
   run "cp public/assets/syndication/tools.js public/static/tools/tools.js" 
```
We couldn't find any references to "public/static/tools" in the code though.

Other than that we tested this and the code is being served correctly. We added a comment to the code to verify it's being served from frontend and not public_website.

### DIFF
--- a/app/assets/javascripts/syndication/resizer.js.coffee
+++ b/app/assets/javascripts/syndication/resizer.js.coffee
@@ -1,0 +1,18 @@
+window.handleMASResizeMessage = (event) ->
+  try
+    if typeof (event.data) is "string"
+      parts = event.data.split("-")
+      prefix = parts[0]
+      height = parseInt(parts[1])
+      if prefix is "MASRESIZE"
+        iframe = document.querySelector(".mas-widget-iframe")
+        iframe.style.height = height + "px"
+    return
+  catch error
+    # Be silent
+
+if window.addEventListener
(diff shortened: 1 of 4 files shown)


paulanthonywilson:
Fair enough. :+1:

benbarnett:
Good spot! Should we have a corresponding PR on public_website that removes these files just to avoid confusion?

paulanthonywilson:
Good point @benbarnett . Maybe that should wait until this change is in production, though.

benbarnett:
Be good to see the PRs though, we can easily coordinate the deployments/merging.

Otherwise :+1:

paulanthonywilson:
Sure - with a 'DO NOT MERGE UNTIL' warning.

benbarnett:
Exactly.

picandocodigo:
Thanks @benbarnett, we've created the new Pull Request for deleting these files on public_website.

benbarnett:
Nice one :+1:

PR #950, opened 2015-03-17
Making latest tools_<generated_asset_hash>.js available as tools.js

JiggyPete:
We want to stop serving tools.js from public website and move this to frontend. 
All the embedded partner-tools make use of /assets/syndication/tools.js
We need to make this a seamless migration.

This PR creates "tools.js" with a kind of 'after' hook on assets:precompile.

We are unsure if we need to update the scripts repo as the tools.js from public_website is currently zipped up as 'static-content' and served by nginx. Ideally we want to be using the new version from frontend.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@ include FileUtils
 
 Rails.application.load_tasks
 
-Rake::Task["assets:precompile"].enhance do
+Rake::Task['assets:precompile'].enhance do
   Rake::Task[:copy_precompiled_tools_js].invoke
 end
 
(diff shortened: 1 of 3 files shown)


neoeno:
Does the asset manifest.json have any bearing here? I don't know very much about it, but I do know it usually contains a list of all the precompiled assets.

JiggyPete:
This is the first time I've heard of manifest.json?

neoeno:
It's usually generated on asset precompile and rails uses it for some reasons I learned about about six months ago but have since forgotten. 

In frontend, if you run a precompile and then `ls public/a/manifest*` you should see a file there.

JiggyPete:
Yeah the manifest includes syndication/tools.js

JiggyPete:
Thanks

neoeno:
Neat! Rails must do it after this task then.

neoeno:
Pushed some style stuff. Apparently `File.exists?` is deprecated in favour of `File.exist?`

picandocodigo:
Updated the Pull Request with a variable instead of console.log.

mconnell:
I had a question about the local variable instead of a global, but after discussion with @JiggyPete and @picandocodigo, they are just declaring it so when they view the source it can be found. It doesn't serve any purpose other than that so :+1:

picandocodigo:
@benbarnett sent his :+1: via Slack, so merging.

PR #953, opened 2015-03-24
Mount debt test

Guntrisoft:


Guntrisoft:
@Tr4pSt3R @darokel @asmega

Tr4pSt3R:
:+1:

app/controllers/debt_test_controller.rb
```diff
@@ -0,0 +1,10 @@
+class DebtTestController < EmbeddedToolsController
+  def parent_template
+    'layouts/engine_unconstrained'
+  end
+
+  protected
+    def category_id
+      ''
```
asmega: do we know what this should be?

Gemfile
```diff
@@ -46,6 +46,7 @@ gem 'budget_planner', '~> 4.0.1.689'
 gem 'car_cost_tool', '~> 1.0.4.110'
 gem 'debt_advice_locator', '~> 2.0'
 gem 'debt_free_day_calculator', '~> 2.1.0'
+gem 'debt_test', git: 'git@github.com:moneyadviceservice/debt_test.git'
```
asmega: hopefully we can fix the build and point to the gem

app/controllers/debt_test_controller.rb
Tr4pSt3R: I will refer that to mani.

Gemfile
darokel: The build is green, let's point to the gem

PR #957, opened 2015-03-26
ensure that correct visited blue is used

Guntrisoft:
and headline is black if it's not a link


Guntrisoft:
@asmega @stevoland

asmega:
:+1:

stevoland:
:+1:

PR #972, opened 2015-03-31
[update] mortgage calculator to -v 1.3.7.625

Tr4pSt3R:


Gemfile.lock
```diff
@@ -678,7 +678,7 @@ DEPENDENCIES
   cream
   csslint_ruby
   cucumber-rails
-  cutback_calculator (~> 0.9.15)
+  cutback_calculator
```
misaka: Is this meant to be un-pinned?

Tr4pSt3R: Yeah, until it's finally released. It has not been released yet

PR #986, opened 2015-04-06
add setter for bug snag key

mconnell:
This accompanies PR: https://github.com/moneyadviceservice/scripts/pull/207 to to reintroduce the bugsnag api key if we want it back.

NOTE: The other commit will need to be merged into scripts before this one goes in.


asmega:
:+1:

paulanthonywilson:
:+1:

PR #988, opened 2015-04-08
mount baby_cost_calculator behind feature toggle

asmega:


mconnell:
Pulled this down and tried it locally. The engine has a couple of js errors and some styling is off but the tool correctly mounts to where it says it. Only query is about the gem pointing to a git url instead of a versioned gem?

Gemfile
```diff
@@ -43,6 +43,7 @@ gem 'uglifier', '>= 1.3.0'
 gem 'action_plans', '~> 3.0.0'
 gem 'advice_plans', '~> 3.1.0'
 gem 'agreements', '~> 2.0.1'
+gem 'baby_cost_calculator', git: 'git@github.com:moneyadviceservice/baby_cost_calculator.git'
```
benlovell: You can use `github: 'moneyadviceservice/baby_cost_calculator'` instead.

spec/requests/tool_integration/baby_cost_calculator_spec.rb
```diff
@@ -0,0 +1,14 @@
+RSpec.describe 'Baby Cost Calculator', type: :request, features: [:baby_cost_calculator] do
```
benlovell: I don't think you need to wrap `features` in an array. The provided feature arguments get normalised along the way.

asmega: Cool, I'll try that next tome

Gemfile
asmega: yes. I will switch to the gem once I get the thing in a better state

mconnell: :+1:

asmega:
@mconnell It's still in a work in progress but in a workable state. I've yet to make a Go pipeline and gem build but plan to do that once I get the thing towards a releasable state.

Styling should be good now. What js errors were you getting?

mconnell:
On loading the app:
GET http://localhost:5000/en/tools/baby-cost-calculator/baby_cost_calculator/data/tips.json 501 (Not Implemented)

The boxes on the right hand side that need to be checked, I been getting:
babytool.js?body=1:539 Uncaught TypeError: undefined is not a function

Also seeing a bind on links that stops being able to switch between languages and clicking on the mas logo loads the homepage content in the same page. https://www.dropbox.com/s/3vvdzhhwbj7t8v8/Screenshot%202015-04-10%2010.37.37.png?dl=0

asmega:
ah yes I remember now. The dummy app works fine. But there are integration issues as you are obviously experiencing. so the plan (@Guntrisoft gets all the credit) was to iframe it to prevent the css cascading issues (which should also fix the js issues too)

It's not responsive anyway so we didn't see this as a problem

darokel:
Can we hold off merging this please. This needs a little more business discussion

asmega:
I'm no longer working on this and someone else has picked this up so closing

PR #990, opened 2015-04-09
Fix RIO homepage in IE8.

neoeno:
* `retirement_shop_(en|cy).png` was bigger than the element, and `background-size` wasn't constraining it in IE8. We've added a smaller version for IE8 and used that.
* The above background image was shifted too far to the left in IE8, obscuring the text. We've shifted it right in IE8. (+1 squashed commit)

@ourmaninamsterdam: I understand that any CSS in a `respond-to` block was excluded from IE8, but I didn't see it work that way. This is the IE8 render after this PR:

image

And this is the IE8 render with [this line](https://github.com/moneyadviceservice/frontend/compare/fix_homepage?expand=1#diff-fec5cbae57b77506827ee2931957db72R30) commented out:

image

That should be a no-op, shouldn't it?


neoeno:
Just spotted an issue with this on mobile. One sec...

neoeno:
Fixed.

ourmaninamsterdam:
Great! 

TBH I cannot think of a more elegant way of doing the image swap. The only other alternative would be to target the `.ie8` class, but that would output the CSS in the stylesheet, whereas iffing `$responsive` omits it completely, if not required.

So :+1: :+1:

alexwllms:
Looks good to me. :+1:

PR #1003, opened 2015-04-17
Corporate Page

darokel:
PR to revamp the corporate section of the website. In summary:

- New corporate home page accessible at `/en/corporate`
- Modifies some core files `category.rb`, `category_tree_reader.rb`, couple of controllers (mainly corporate) and decorators to accommodate corporate specific categories. 
- Work was previously done by Doug and finished by me hence the different commits. 

Still need to squash and clean. 

@asmega


app/controllers/corporate_controller.rb
```diff
@@ -1,9 +1,27 @@
...
+    Core::CategoryTreeReader.new.call(categories)
+  end
+
+  def assign_categories
+    @category = Core::CategoryReader.new('corporate-home').call
+    assign_active_categories(@category)
+    # @other = Core::Registry::Repository[:category].find('corporate')
+    # @categories = CategoryNavigationDecorator.decorate_collection(
+    #   category_tree(@other['contents']).children
+    # )
```
asmega: still need these comments?

app/views/corporate/extra/_our_debt_work_corporate.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="l-inpage-nav">
+  <div class="l-inpage-nav__item">
+    <ul class="link-list-primary link-list-primary--no-border">
+      <li class="link-list-primary__item">
+        <a href="#" class="link-list-primary__text is-selected">Lorem ipsum</a>
```
asmega: still awaiting copy?

app/controllers/corporate_controller.rb
darokel: Good spot

app/views/corporate/extra/_our_debt_work_corporate.html.erb
darokel: I don't actually know what these partials are used for. @ourmaninamsterdam any idea?

ourmaninamsterdam: I think these exist from when Doug brought all the separate corporate branches under a single branch.

dougdroper: correct, you can remove them :)

ourmaninamsterdam: :+1:

app/controllers/concerns/navigation.rb
```diff
@@ -11,7 +11,7 @@ module Navigation
   attr_reader :active_categories
 
   def assign_active_categories(*categories)
-    categories.each do |category|
+    categories.map do |category|
```
mconnell: I can't see where the change from `.each` to `.map` is used?

darokel: Not too sure myself, this was originally done by @dougdroper. Looks like he wanted `assign_active_categories` to explicitly return a value?

mconnell: That's what it looks like, but I couldn't see it being assigned or used anywhere like that.

app/controllers/application_controller.rb
```diff
@@ -93,12 +93,29 @@ def check_syndicated_layout
...
-    @category_navigation ||= CategoryNavigationDecorator.decorate_collection(category_tree.children)
+  def navigation_categories
+    Core::Registry::Repository[:category].all
+  end
+
+  def corporate_categories
+    Core::Registry::Repository[:category].find('corporate-categories')['contents']
+  end
+
+  def corporate_category?(category, corporate = corporate_categories, categories = [])
```
mconnell: This method is listed as a private method, but is also declared as a helper_method. Making it private to the class but publicly available as a view helper. Intentional?

The method is doing a number of different things (including being recursive) which I would be a bit uncomfortable changing without having tests describing the existing behaviour.

darokel: Again, not sure what the thought process was here.

asmega:
:+1:

ourmaninamsterdam:
:+1:

PR #1007, opened 2015-04-21
Remove guard and associated files

alexwllms:
After a short Slack poll it seems no one uses guard, so I have removed it as part of general housekeeping. Keeping the Gemfile clean!

If you would like it kept then speak!

@moneyadviceservice/frontend 

@benlovell @darokel


benlovell:
ALL THE GEMS! :boom: :bomb: :ship:

Guntrisoft:
:+1:

PR #1008, opened 2015-04-22
Fix issue where chat button isn't present on a MAS page.

Guntrisoft:
No need for the javascript to kick in for the button if it isn't there

This was causing the following JS error on the /en/campaigns/debt-management page (which doesn't have the usual footer).

Uncaught TypeError: Cannot set property 'innerHTML' of null


Guntrisoft:
@ourmaninamsterdam @benbarnett

ourmaninamsterdam:
:+1:

benbarnett:
:+1:

PR #1010, opened 2015-04-22
Corporate feature

darokel:
Places the corporate page behind a feature flag.

### DIFF
--- a/features/corporate_page.feature
+++ b/features/corporate_page.feature
@@ -1,0 +1,13 @@
+Feature: Settings
+  As a corporate partner
+  I want to be able to view more information on partnering
+  So that I am aware of different ways of partnering with MAS
+
+  @enable-corporate
+  Scenario: Accessing the corporate page
+    When  I visit the corporate page
+    Then  I should be informed that I am on the correct page
+
+  Scenario: Corporate page is not accessible when feature is disabled
+    When I visit the corporate page
+    Then  I am told that the functionality is not implemented
--- a/features/support/ui/pages/corporate.rb
+++ b/features/support/ui/pages/corporate.rb
@@ -3,6 +3,5 @@ require_relative '../page'
 module UI::Pages
   class Corporate < UI::Page
     set_url '{/locale}/corporate'
-
   end
 end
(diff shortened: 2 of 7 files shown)


asmega:
:+1:

features/corporate_page.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Settings
+  As a corporate partner
+  I want to be able to view more information on partnering
+  so that I am aware of different ways of partnering with MAS.
```
benlovell: Capitalise `so` and the full stop can go to I reckon.

darokel:
@benlovell All good?

features/corporate_page.feature
```diff
@@ -0,0 +1,13 @@
...
+
+  @enable-corporate
+  Scenario: Accessing the corporate page
+    When  I visit the corporate page
+    Then  I should be informed that I am on the correct page
+
+  Scenario: Corporate page is not accessible when feature is disabled
+    When I visit the corporate page
```
benlovell: If we are going to align the non-gherkin reserved stuff (which I don't necessarily agree with myself, but it seems other folks like to), then it should be consistent. This one is out a little.

features/support/ui/pages/corporate.rb
```diff
@@ -0,0 +1,8 @@
+require_relative '../page'
+
+module UI::Pages
+  class Corporate < UI::Page
+    set_url '{/locale}/corporate'
+
```
benlovell: Whoops.

spec/controllers/corporate_controller_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe CorporateController, type: :controller do
+RSpec.describe CorporateController, type: :controller, features: [:corporate] do
```
benlovell: I'm certain you don't need to wrap `:corporate` in an array. They get normalised in the library.

darokel: There are a couple of other files which follow the same format. I personally don't mind whether we use arrays or not (for single features) but I'd like consistency. I suppose having single features in an array clarify more intently that multiple features can be passed in. What do you reckon?

features/support/ui/pages/corporate.rb
darokel: Dang!

PR #1011, opened 2015-04-22
Check in corporate cassettes

darokel:
These cassettes are needed for the corporate page related unit/feature specs.


asmega:
did you figure out why `Partners` is capitalised? as all the other aren't. Will this cause a problem?

darokel:
Not sure. It's the same way in qa and production. Hasn't caused a problem yet and I don't think there are plans to change it

asmega:
:+1:

PR #1012, opened 2015-04-22
contact us and accessibility pages use contento

asmega:
I tried to make this backwards compatible but it turns out that the `Feature` library isn't available when setting up core to use different drivers/repositories


darokel:
:+1:

PR #1013, opened 2015-04-23
Fix issue when no child categories are present

darokel:
Fixes recent issue of category pages blowing up and returning 500s. Incident report can be found here (if you're a MAS employee) https://moneyadviceserviceuk.atlassian.net/wiki/display/DEV/2015-04-23+-+Site+Category+Pages+Down.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -108,7 +108,9 @@ class ApplicationController < ActionController::Base
   def corporate_category?(category, corporate = corporate_categories, categories = [])
     categories << corporate.map {|c| c['id']}
     return true if categories.flatten.include?(category)
-    corporate_category?(category, corporate.first['contents'], categories.flatten) if corporate.first['contents']
+    unless corporate.first['contents']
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -108,7 +108,9 @@ def corporate_categories
   def corporate_category?(category, corporate = corporate_categories, categories = [])
     categories << corporate.map {|c| c['id']}
     return true if categories.flatten.include?(category)
-    corporate_category?(category, corporate.first['contents'], categories.flatten) if corporate.first['contents']
+    unless corporate.first['contents']
```
asmega: how does this fix the problem?

darokel: Checks for empty child categories first before trying to run stuff on `nil`.

asmega: ah I think I remember now if has empty contents if it has children.
would `if corporate.has_children?` not be better?

darokel: It would be, but right here corporate is just an `Array` that's put together from another method.

asmega:
:+1:

PR #1014, opened 2015-04-23
add sitemap

asmega:
public website version pulls from old cms hippo. This one pulls from the new cms contento


asmega:
how about now?

darokel:
Much better :+1:

PR #1015, opened 2015-04-24
footer links go to contento pages/categories

asmega:


darokel:
Just one comment but apart from that :+1:

PR #1020, opened 2015-04-29
separating 'More In' and 'Related Links' into two different components

Guntrisoft:
so that they can eventually be styled differently.

This needs to be done, as the related links on the right hand side of an article are going to be styled differently from the 'More in' list of links, in another PR

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_more_in.scss
+++ b/app/assets/stylesheets/components/page_specific/_more_in.scss
@@ -1,0 +1,17 @@
+// Content box that appears on article pages displaying links to other content in the current category
+//
+// Styleguide more in
+
+.more-in__heading {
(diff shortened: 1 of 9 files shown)


Guntrisoft:
@benbarnett @alexwllms - could you take a look at this - I'm a bit nervous about this change as I hope I've understood fully about how the related links stuff works.

The tests pass ok!

benbarnett:
Seems fine, but might be worth changing the data in the fixture for one of them – its a bit confusing in the styleguide at the moment as they look identical.

Guntrisoft:
@benbarnett done that now

benbarnett:
Cool :+1:

alexwllms:
LGTM :tada:

PR #1025, opened 2015-04-29
Update features.yml.sample with production conf

tomas-stefano:


Tr4pSt3R:
hey @tomas-stefano you could make it easier to read by ordering them alphabetically

tomas-stefano:
Good catch. I just copy from the scripts repo. Both could be using that rule :)

asmega:
:+1:

Tr4pSt3R:
:+1:

PR #1035, opened 2015-05-04
Retirement landing pages copy changes

neoeno:
Feedback on the approach in https://github.com/moneyadviceservice/frontend/commit/9d4ebcc00db7a65058c557373eb11d0454256e6d would be very much appreciated! I wanted to enable the locale switcher in the nav bar but there didn't seem to be an established way for landing pages.


neoeno:
@darokel Did as you suggested to see how it panned out — @paulanthonywilson and I looked over the result, and extracting the constants feels better but I'm struggling to convince myself it should be a concern. Then again, I can't think of anywhere better to put it. It's certainly not _employing_ `ActiveSupport::Concern`.

Maybe I'll create `lib/landing_pages.rb` and generalise it to contain things like this in future. Wyt?

neoeno:
@paulanthonywilson wyt of this approach? https://github.com/moneyadviceservice/frontend/commit/a8c10f4660746ab1f2b6f1b493dd88165e362886

darokel:
Generally you would put any code that is not specific to the domain of the application into the lib folder. I'm not sure whether it's the best place to put landing page helpers. Just out of curiosity why were you trying to `extend ActiveSupport::Concern`? 

Alternatively, why not just create a plain old module in `models/concerns`. At least that way, when people see it at the controller level, they'd naturally check the models directory?

darokel:
Ignore my previous comment. On first glance, I was a bit skeptical but I actually like this approach. How extendable is the `LandingPagePaths`, could we re-use it for other landing pages in the future?

neoeno:
Yep — the process would just be to add the URLs to `config/locales/landing_pages/paths.*.yml` and then use `LandingPagePaths` in the same way as it's used in the routes/controller in this PR.

There's nothing about it that makes it landing page specific either — but I figured it could probably be repurposed later if it found wider use.

darokel:
Looks good :+1: :tada:

paulanthonywilson:
:+1: fine.

neoeno:
Thanks both!

PR #1036, opened 2015-05-07
Fix wrong feature flag in RetirementsController spec.

neoeno:
This didn't show up before because the spec file didn't have `_spec.rb` on the end.

Then I renamed it in https://github.com/moneyadviceservice/frontend/pull/1035, but only tested it with the feature enabled in my local repo. Maybe in tests all features could be disabled by default, any thoughts?

@paulanthonywilson @munckymagik


paulanthonywilson:
:+1:

munckymagik:
:+1:

PR #1037, opened 2015-05-07
Remove public website cms fallbacks

asmega:
- We no longer depend on public website for cms content so I am removing the fallbacks.
- This requires regenerating all the vcr cassettes and some test tweaks


tomas-stefano:
Great step. :+1: :+1: :shipit: :ship:

PR #1038, opened 2015-05-07
Remove public website cms fallbacks

asmega:
on a 404 it will now return nil which is what the public wesbite fallback
repository used to do


darokel:
:+1: :tada:

tomas-stefano:
:+1: :+1: :ship:

PR #1039, opened 2015-05-07
Put Welsh retirement landing pages behind feature flag.

neoeno:
I think this is going to take longer than I hoped to send live — this PR is so I don't block any of youse getting to production

@baphled @Tr4pSt3R since I think you're on the path to deployment

### DIFF
--- a/app/controllers/retirements_controller.rb
+++ b/app/controllers/retirements_controller.rb
@@ -3,7 +3,16 @@ class RetirementsController < ApplicationController
   helper_method :locale_options
 
   def locale_options
-    LandingPagePaths.locale_options(params[:controller], params[:action])
+    if Feature.active? :welsh_retirement_landing_pages
+      LandingPagePaths.locale_options(params[:controller], params[:action])
+    else
+      []
+    end
+  end
+
+  def alternate_locales
+    return if Feature.active? :welsh_retirement_landing_pages
+    []
   end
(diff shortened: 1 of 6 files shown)


neoeno:
Added another sneaky locales commit.

munckymagik:
:+1:

darokel:
Could you add this flag to the `features.yml.sample` too?

neoeno:
@darokel Can do. What's the purpose of `features.yml.sample`? Are we supposed to have all feature flags enabled there?

darokel:
@neoeno Because we don't check in the actual file, it's supposed to be a sourth of truth for all the features the application requires to run certain tools/sections. There's a command in the readme `cp config/features.yml.sample config/features.yml` which would copy over the most up to date features list to your local file. Not sure whether the values should be set to true by default though, this may bug out some specs.

neoeno:
@darokel OK — noted & committed

baphled:
Who do I bill for reading these ;D

neoeno:
Ha, oops — sorry @baphled. Who was I trying to tag... can't remember...

neoeno:
@asmega probably!

baphled:
Haha @caden, s'all good, it's nice to be wanted :)

mconnell:
:+1:

PR #1042, opened 2015-05-11
cms data can be html or markdown

asmega:
This adds a test to ensure frontend is capable of handling data from the cms that is in markdown or html (the cms has already processed the markdown to html)


mconnell:
:+1:

tomas-stefano:
LGTM :+1: :ship:

PR #1043, opened 2015-05-11
remove definition of debt-test parent template

Tr4pSt3R:
* setting the parent layout in this way breaks setting the syndicated
  * layout when it is need.
  * As far as I know, setting the layout makes no difference to the layout

### DIFF
--- a/app/controllers/debt_test_controller.rb
+++ b/app/controllers/debt_test_controller.rb
@@ -1,6 +1,10 @@
 class DebtTestController < EmbeddedToolsController
   def parent_template
-    'layouts/engine_unconstrained'
+    if syndicated_tool_request?
+      'layouts/engine_syndicated'
+    else
+      'layouts/engine_unconstrained'
+    end
   end
 
   protected


tomas-stefano:
:+1: @Guntrisoft  is there any side effects of this removal?

Tr4pSt3R:
I have investigated it and nothing apparent is missing. Have look please @maniarasant && @Guntrisoft

benbarnett:
I think although there may be very little changes right now, if the engines set themselves to use these templates it means we can change them later without having to modify the main template. What do you think?

Tr4pSt3R:
Thanks @benbarnett; I will have further chats about it with @asmega and @Guntrisoft who appeared to have worked on it

Guntrisoft:
@Tr4pSt3R yes there are side effects of changing this. The design for the question pages was required to have a 'full bleed' grey going across the page - changing back to engine.html.erb adds a layout div which contrains all the content to a certain max-width, resulting in white space either side of the grey strip.

Might need to go through this in more detail, to work out what solution we can come up with, which doesn't affect the layout of the tool.

benbarnett:
:+1:

Guntrisoft:
:+1:

PR #1046, opened 2015-05-13
Remove welsh retirement landing pages feature flag

neoeno:
It's going live, and it shouldn't need to be conditionally enabled or disabled.

### DIFF
--- a/app/controllers/retirements_controller.rb
+++ b/app/controllers/retirements_controller.rb
@@ -3,16 +3,7 @@ class RetirementsController < ApplicationController
   helper_method :locale_options
 
   def locale_options
-    if Feature.active? :welsh_retirement_landing_pages
-      LandingPagePaths.locale_options(params[:controller], params[:action])
-    else
-      []
-    end
-  end
-
-  def alternate_locales
-    return if Feature.active? :welsh_retirement_landing_pages
-    []
+    LandingPagePaths.locale_options(params[:controller], params[:action])
   end
 
   def display_menu_button_in_header?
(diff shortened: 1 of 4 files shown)


app/controllers/retirements_controller.rb
```diff
@@ -3,16 +3,7 @@ class RetirementsController < ApplicationController
...
 
   def locale_options
-    if Feature.active? :welsh_retirement_landing_pages
-      LandingPagePaths.locale_options(params[:controller], params[:action])
-    else
-      []
-    end
-  end
-
-  def alternate_locales
```
asmega: did you want to remove this method?

config/routes.rb
```diff
@@ -112,12 +112,9 @@ def not_implemented
 
     Feature.with(:pensions_and_retirement) do
       get LandingPagePaths.path(:retirements, :index, :en),     to: 'retirements#index'
+      get LandingPagePaths.path(:retirements, :index, :cy),     to: 'retirements#index'
       get LandingPagePaths.path(:retirements, :budgeting, :en), to: 'retirements#budgeting'
-
-      Feature.with(:welsh_retirement_landing_pages) do
-        get LandingPagePaths.path(:retirements, :index, :cy),     to: 'retirements#index'
```
asmega: a route was removed. is that desired?

app/controllers/retirements_controller.rb
neoeno: I did — `locale_options` is sufficient for the intended behaviour.

config/routes.rb
neoeno: It's still there — just moved out of the feature block. See: https://github.com/moneyadviceservice/frontend/pull/1046/files#diff-21497849d8f00507c9c8dcaf6288b136R115

asmega:
:+1:

paulanthonywilson:
:+1:

PR #1049, opened 2015-05-13
TP-6352: Sticky column - back to top edge case bug

Guntrisoft:
- Sometimes when back to top is clicked, the browser gets
  confused, thinking that the sticky component is at the bottom.
- Added check to ensure that it returns to it's natural static
  position when back to top is pressed

### DIFF
--- a/app/assets/javascripts/components/StickyColumn.js
+++ b/app/assets/javascripts/components/StickyColumn.js
@@ -148,12 +148,20 @@ define(['jquery', 'DoughBaseComponent', 'eventsWithPromises'], function($, Dough
       }
     }
 
-    if (this.isAtBottom && scrollTop < this.bottom) {
+    if (this.isAtBottom && scrollTop < this.bottom && scrollTop > this.top) {
       this.$el
         .addClass(this.config.classes.fixed)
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/StickyColumn.js
```diff
@@ -148,13 +148,21 @@ define(['jquery', 'DoughBaseComponent', 'eventsWithPromises'], function($, Dough
       }
     }
 
-    if (this.isAtBottom && scrollTop < this.bottom) {
+    if (this.isAtBottom && scrollTop < this.bottom && scrollTop > this.top) {
```
benbarnett: Do you think nesting these two `if (this.isAtBottom)` conditions in a single one would make it easier to read? I.e. how you've done it for `if (this.isFixed)` here https://github.com/moneyadviceservice/frontend/blob/sticky_column_back_top_fix/app/assets/javascripts/components/StickyColumn.js#L134

Guntrisoft: fair - will change

benbarnett:
:+1:

jonalport:
Good job, Guntrisoft! :+1:

PR #1050, opened 2015-05-13
TP-6278 - Change syndicate redundancy url

tomas-stefano:


asmega:
:+1: you may want to update the test harness after this change gets merged

tomas-stefano:
Thanks for the heads up. Already done it.

PR #926, opened 2015-02-23
Revert "Ignore .env and replacing with .env.dev"

tomas-stefano:
* The application CI depends on it. Revert now in order to see this in a more
deep investigation.

This reverts commit 2d22c1c92125e846461d32ced9367252a50c77cd.

PR #924, opened 2015-02-23
Environment variables

tomas-stefano:
* Make sure that .env is ignored.
* Add instruction about Copying the .env.dev file
* Add instructions to Readme about CMS

PR #1053, opened 2015-05-14
in .env point to QA cms

asmega:
this is so developers who want to fire up frontend will have access to
dummy data. reducing overhead in setup. developers can choose to set
this back to their local version if needed

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 MAS_PUBLIC_WEBSITE_URL=https://www.moneyadviceservice.org.uk/:locale/
-MAS_CMS_URL=http://localhost:3000/
+MAS_CMS_URL=http://comfy.qa.mas
 MAS_CONTENT_SERVICE_URL=http://localhost:8080/content-service/
 GOOGLE_API_KEY=AIzaSyCydfZlpTZlIUUUSbtab5t0oOE-y1GzHy0
 GOOGLE_API_CX_EN=011931388747222259444:razn90vopwc
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+MAS_CMS_URL = "http://localhost:3000/"
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ $ foreman s
 
 ### Change CMS URL Path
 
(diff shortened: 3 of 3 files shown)


jonalport:
:+1:

neoeno:
Yeah this always catches me out. :+1:

Might want to update the `README.md` [here](https://github.com/moneyadviceservice/frontend/blame/master/README.md#L62)

paulanthonywilson:
Oh, yeah. That should save pain :+1: :+1:

paulanthonywilson:
Thought on this, though - should this be .env.example and add .env to .gitignore?

tomas-stefano:
Yes, I tried in a very old PR  to remove .env but a lot of tests fails due to the expectation of these variables. I will search for the PR.

tomas-stefano:
This one: https://github.com/moneyadviceservice/frontend/pull/924

tomas-stefano:
And then reverting: https://github.com/moneyadviceservice/frontend/pull/926 hehe

darokel:
Can't believe I didn't even notice `.env` was checked in. This shouldn't be in here.

darokel:
@tomas-stefano Where did those tests fail? On the CI?

tomas-stefano:
On CI and locally. Try to remove .env and runs the specs.

paulanthonywilson:
bummer. There’s probably a way round it, but will take someone’s time.

tomas-stefano:
Yes, exactly and time in the 'old world' was difficult to have. Then I decide to revert.

darokel:
@tomas-stefano Yeah they would fail locally if that file isn't there but on the CI we would need to make those keys available within the `scripts` repo. I can take a look at this tomorrow hopefully

tomas-stefano:
By the way, update the README and this PR :+1: But we need to make another PR ignoring this file from git repo.

tomas-stefano:
@darokel :+1:

asmega:
I've made a change and added a `.env.test` file as VCR uses hostname matching in order for the cassettes to be loaded. Otherwise tests will fall over

PR #1055, opened 2015-05-14
Conditional loading of syndicated tool iframe resizer 

benbarnett:
Some tools do not have any benefit from having the iframe resizer when being syndicated. For the debt test tool, it was causing problems with the scrolling. This enables tools to disable the iframe resizer if they need.

### DIFF
--- a/app/controllers/debt_test_controller.rb
+++ b/app/controllers/debt_test_controller.rb
@@ -7,6 +7,10 @@ class DebtTestController < EmbeddedToolsController
     end
   end
 
+  def exclude_syndicated_iframe_resizer?
+    syndicated_tool_request?
+  end
+
   protected
     def category_id
       'before-you-borrow'
--- a/app/views/layouts/engine_syndicated.html.erb
+++ b/app/views/layouts/engine_syndicated.html.erb
@@ -15,10 +15,12 @@
     </div>
   </div>
   <% content_for(:javascripts) do %>
-    <%= javascript_include_tag "syndication/iframeResizer" %>
-    <script>
-      window.iframeResizer('MASRESIZE-').start();
-    </script>
+    <% unless exclude_syndicated_iframe_resizer? %>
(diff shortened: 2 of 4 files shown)


app/controllers/debt_test_controller.rb
```diff
@@ -7,6 +7,10 @@ def parent_template
     end
   end
 
+  def exclude_syndicated_iframe_resizer?
+    return true if syndicated_tool_request?
```
asmega: just thought that `syndicated_tool_request?` will work don't need the `return true` bit

benbarnett: oh yeah good point

tomas-stefano:
Cool. LGTM :+1:

app/controllers/debt_test_controller.rb
tomas-stefano: Good point :)

rnalexander:
:+1:  :shipit:

paulanthonywilson:
:+1:

asmega:
:+1:

benbarnett:
Thanks all

tomas-stefano:
:+1:

app/views/layouts/engine_syndicated.html.erb
```diff
@@ -15,10 +15,12 @@
     </div>
   </div>
   <% content_for(:javascripts) do %>
-    <%= javascript_include_tag "syndication/iframeResizer" %>
-    <script>
-      window.iframeResizer('MASRESIZE-').start();
-    </script>
+    <% unless exclude_syndicated_iframe_resizer? %>
```
JiggyPete: My only suggestion would be to flip the logic and the name....for example
if include_syndicated_iframe_resizer?

other than that :+1:

benbarnett: Yeah it is a bit of a double negative I suppose!

PR #1060, opened 2015-05-20
Features toggle should be sort in alphabtical order

tomas-stefano:
See https://github.com/moneyadviceservice/scripts/pull/239 for more details


darokel:
:+1:

PR #1061, opened 2015-05-20
Better wording for CSS styleguide

benbarnett:
We favour `js-` prefixes for Javascript selectors, whereas the styleguide was suggesting id's as well.


urfolomeus:
Perhaps it would help to also be explicit about not using ids?

benbarnett:
Fair comment. The prefix mentioned further up which the diff doesn't show
but yeah could be more explicit.

urfolomeus:
Oh does it? Sorry. I couldn't see anything above that line that mentioned IDs. :)

mconnell:
:+1:

PR #1075, opened 2015-06-01
Bumping payday loans

benbarnett:


jonalport:
Not sure of the implications of this but syntax is A-OK :+1:

PR #1078, opened 2015-06-02
[WIP] Document clean architecture

tomas-stefano:
I bring this commit from an old/stale branch and I think that is important to be documented on the Readme this clean architecture. Wdyt?

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,25 @@ automatically available to the asset pipeline.
 
 ### Patterns
 
+#### Clean Architecture
+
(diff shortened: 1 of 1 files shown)


darokel:
Good idea! Might be worth double checking whether we need to add more info

tomas-stefano:
Until now, my actual knowledge is the same as this one list on the Readme. I double checked and I think I am not missing anything.

tomas-stefano:
Is this good to go?

darokel:
Not yet. What I meant was that we probably need to add some brief context underneath the heading of the core concepts. Specifically for Entities, Registries and Repositories

PR #1079, opened 2015-06-03
Redirect to root after feedback if no referrer url

paulanthonywilson:
After technical feedback is received the user is returned back to the
referring page. This could cause 500 errors if the url was entered
directly.

This fix redirects the user to the root page if there is no referrer.


pmcostello:
:+1:

neoeno:
:+1:

PR #1083, opened 2015-06-09
Update syndication & tools pages

jonalport:
A handful of small changes, mostly to do with copy.
Detailed in TP ticket #5621

### DIFF
--- a/app/views/corporate/extra/_syndication.html.erb
+++ b/app/views/corporate/extra/_syndication.html.erb
@@ -3,7 +3,7 @@
 <div class="l-promo-partnerships promo promo--partnerships">
   <div class="l-promo-partnerships__item">
     <%= heading_tag(level: 3, class: 'promo__heading') do %>
-      <%= link_to tool_path('pension-calculator'), class: 'promo__link' do %>
+      <%= link_to t('corporate_tool_directory.popular_tools.tool_one_url'), class: 'promo__link' do %>
         <%= image_tag('corporate/pension_calculator.png', class: 'promo__img') %>
         <%= t('corporate_tool_directory.popular_tools.tool_one_heading') %>
       <% end %>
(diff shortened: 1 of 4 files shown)


app/views/corporate/extra/_tool.erb
```diff
@@ -1,3 +1,9 @@
+<h2 id="get-this-tool" role="heading" aria-level="2">
```
benbarnett: There's a `heading_tag` helper we've got that could generate these for you. Example here: https://github.com/moneyadviceservice/frontend/blob/master/app/views/campaigns/_section.html.erb#L4

app/views/corporate/extra/_tool.erb
```diff
@@ -67,6 +73,10 @@
   </div>
 
   <% if flash[:partner] %>
+    <h2 id="your-embed-code" role="heading" aria-level="2">
```
benbarnett: Ditto re heading

benbarnett:
Up to you regarding the headings, so :+1:

darokel:
:+1:

app/views/corporate/extra/_syndication.html.erb
```diff
@@ -3,7 +3,7 @@
 <div class="l-promo-partnerships promo promo--partnerships">
   <div class="l-promo-partnerships__item">
     <%= heading_tag(level: 3, class: 'promo__heading') do %>
-      <%= link_to tool_path('pension-calculator'), class: 'promo__link' do %>
+      <%= link_to t('corporate_tool_directory.popular_tools.tool_one_url'), class: 'promo__link' do %>
```
asmega: can we not have this as an array of hashes that we can iterate over? that way when someone adds a new thing they wont have to copy/paste markup

jonalport: Since there are only 3 allowed in the design, we can can leave it as is I think. But thanks for the suggestion! :)

app/views/corporate/extra/_tool.erb
jonalport: Nice one, have updated

PR #1084, opened 2015-06-09
Corporate partners export auth

darokel:
This locks down `/en/corporate/export-partners` behind http auth. Accompanying PR for the creds are placed in `scripts`. 

https://github.com/moneyadviceservice/scripts/pull/265


features/support/env.rb
```diff
@@ -94,5 +95,11 @@
   create(:corporate_partner)
 end
 
+Before('@auth-required') do
+  allow(Authenticable).to receive(:username)  { ENV['AUTH_USERNAME'] }
+  allow(Authenticable).to receive(:password)  { ENV['AUTH_PASSWORD'] }
```
munckymagik: Oh! Will these methods (username, password) on Authenticable not return the value of these env vars when when testing? What I mean is, it looks like these stubs reimplement the same thing as the real module does, is that not the case?

urfolomeus:
:+1:

features/support/env.rb
darokel: Argh yeah I forgot to replace those with some explicit fake data I was going to pass in. Thanks for spotting this.

munckymagik:
:+1: once the stubbing thing is sorted.

PR #1085, opened 2015-06-15
[WIP] - TP-6344 - Account page - Remove goals section completely

Guntrisoft:
screen shot 2015-06-15 at 15 19 39

_This might not be needed anymore, after an update from Mani - we might end up keeping the Goals section after all - putting this PR here, in case the original subject of this task ends up being what's required!_

- Removing goals from profile/edit template
- Removing backend controller and model code
- Removing translations to do with Goals
- Removing routing to do with Goals
- Removing cucumber features/steps to do with Goals
- Making 'saved tools' section go 100% width on all viewport widths
  now that the space to the left of it has been removed

Removal of the DB columns for goals stuff, will be done in a separate task by Tomas/Jones after they have done a dump of the goals data which is in the DB, to pass onto Jacqui.


darokel:
:+1: May want to double check any associated specs for the user model

Guntrisoft:
This isn't required anymore, as we are keeping the goals section now

PR #1086, opened 2015-06-15
Redirect user to tool id param

jonalport:
...to prevent 404 on tool pages with >2 URL slug segments.

Look OK, @darokel?


asmega:
:+1:

PR #1087, opened 2015-06-15
decrease Rails cache expiry to 1 hour

asmega:
This is so the caching of the categories is reduced to every hour instead of once a day. 
I'm not aware of much else that uses caching.


tomas-stefano:
:+1:

Tr4pSt3R:
:100: :+1:

PR #1090, opened 2015-06-18
TP-6297: Updating dough-ruby gem version

Guntrisoft:


benbarnett:
:+1:

jonalport:
:+1:

Guntrisoft:
strange - I only did bundle update dough-ruby ...?

benbarnett:
I just did it `bundle update dough-ruby` on master (before this was merged) and rack got updated too.

neoeno:
`bundle update dough-ruby` is allowed to update all gems that `dough-ruby` depends on. In this case, dough-ruby depends on rails which depends on actionpack which depends on rack.

darokel:
Yeah I generally use `bundle update --source gemname` first. Just tried it now and only `dough-ruby` was updated. Here's some more info on the difference http://ilikestuffblog.com/2012/07/01/you-should-update-one-gem-at-a-time-with-bundler-heres-how/

PR #1091, opened 2015-06-18
TP-6432: Add componentName to existing dough components in frontend

Guntrisoft:
screen shot 2015-06-18 at 10 47 08

ClearInput, Newsletter, StickyColumn etc etc don't have a componentName set in the JS.
 
This is causing DoughBaseComponent to spit out a console warning.
 
This task is to add names to each JS module so this warning goes away.


benbarnett:
:+1:

jonalport:
:+1:

PR #1094, opened 2015-06-19
Update global styles (#6288)

jonalport:
This updates the font sizing and associated spacing on headings through the site. Have just separated it out from the category styling branch.

### DIFF
--- a/app/assets/stylesheets/_enhanced.scss
+++ b/app/assets/stylesheets/_enhanced.scss
@@ -1,20 +1,37 @@
-@import "settings";
+@import 'settings';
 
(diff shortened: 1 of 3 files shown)


asmega:
:+1:

benbarnett:
Much better, am definitely pro-listing out imports rather than globbing. :+1:

Guntrisoft:
Just pointed out to @jonalport that this change has a great knock on effect to all tools across the site, and each one will probably need to be reviewed by UX/Design to ensure they are happy with how they look now these font sizes have been implemented.

PR #1095, opened 2015-06-22
TP-6317 TP-6344: Profile page changes

Guntrisoft:
screen shot 2015-06-22 at 13 50 45

- swap the goals and saved tools around
- move the recommended tools section directly
  below the saved tools
- drop down to only two recommended tools
- replace the car cost tool with retirement tool


benbarnett:
:+1:

Tr4pSt3R:
:+1: :100:

PR #1099, opened 2015-06-24
Partners embed code submission

darokel:
Allows partners to re-request for another embed code using an existing email address. 

mysizq4kyo

### DIFF
--- a/app/controllers/corporate_controller.rb
+++ b/app/controllers/corporate_controller.rb
@@ -18,7 +18,8 @@ class CorporateController < ArticlesController
   end
 
   def create
-    @partner = CorporatePartner.new(partner_params)
+    @partner = CorporatePartner.find_or_initialize_by(params[:email])
+    @partner.assign_attributes(partner_params)
     @partner.tool_name = tool_friendly_name
 
     if @partner.save
--- a/app/controllers/corporate_controller.rb
+++ b/app/controllers/corporate_controller.rb
@@ -18,7 +18,7 @@ class CorporateController < ArticlesController
   end
 
   def create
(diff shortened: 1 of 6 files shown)


tomas-stefano:
:+1:

Tr4pSt3R:
:+1:

app/controllers/corporate_controller.rb
```diff
@@ -18,7 +18,8 @@ def show
   end
 
   def create
-    @partner = CorporatePartner.new(partner_params)
+    @partner = CorporatePartner.find_or_initialize_by(params[:email])
```
neoeno: I feel sure I'm missing something here, but I can't google it so I'll ask — is this doing something like:

`CorporatePartner.find_or_initialize_by('email@example.com')`

What's the behaviour of `find_or_initialize_by` with a string rather than a hash?

darokel: Oopsie, good call on that. Resolved here https://github.com/moneyadviceservice/frontend/commit/101682bc11ebb6bd495cfeeb42c8abc6f96198c9. So `find_or_initialize_by` does expect a hash of attributes but it still somehow worked when passed a string..weird

neoeno: That is odd.

I noticed in your commit that you changed `params[:email]` to `params[:corporate_partner][:email]` — which would suggest that `params[:email]` might have returned `nil`.

`Model.find_or_initialize_by(nil)` will return the first record in `Model` — so good we caught that! The reason I go into detail is — maybe this is a sign the tests here need tweaking to make thm more observant?

PR #1101, opened 2015-06-24
TP-6444: Remove menu button from profile and registration pages

Guntrisoft:
screen shot 2015-06-24 at 14 56 03

- Added method to three controllers to remove the menu button
  as there is no menu markup output on these sections of pages
- This was causing a "t.find is not a function" JS error appearing no these pages.

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -1,6 +1,8 @@
 class CampaignsController < ApplicationController
   decorates_assigned :campaign, with: CampaignPage::CampaignDecorator
 
+  include SuppressMenuButton
+
   def show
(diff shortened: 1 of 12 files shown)


benbarnett:
I think this also belongs on couple of other pages: https://www.moneyadviceservice.org.uk/en/campaigns/debt-management (and I think all other campaign pages actually)

Guntrisoft:
@benbarnett campaigns controller already suppresses the button, but I should change the campaigns controller to include the new 'concerns' module, instead of repeating the method again

Guntrisoft:
Have included the new suppress menu button concern module in lots of other controllers which overrode this method

benbarnett:
Yeah that's what I was getting at (I think!). Nice one :+1:

PR #1107, opened 2015-06-30
Styling button element for Dough collapsible

benbarnett:
The button was missing a hover state, which highlighted the fact there were two hit areas. This fixes this, and adds the hover states.

The diff looks more serious than the change actually was, its really just a few styles moving around.


asmega:
:+1:

munckymagik:
:+1:

urfolomeus:
:1:

PR #1112, opened 2015-07-02
Renaming page_specific link_list -> category_link_list

benbarnett:
Naming clash that was spilling to pages using link_list from common. Mainly the pensions and retirement landing page.

Also adds `@extend .unstyled-list` as this was because inherited from the 'common' component, and now we've separated them it needs to be declared here, too.


benbarnett:
Link to affected pensions page that this change fixes here: https://www.moneyadviceservice.org.uk/en/pensions-and-retirement

asmega:
:+1:

Guntrisoft:
:+1:

PR #1119, opened 2015-07-06
id:6391 comment: design tweeks from Irina after feedback

Guntrisoft:
- Search text increased, after Irina felt it was too small (14px -> 16px)
- padding tweaks and font-size change of copy


benbarnett:
:+1:

darokel:
:+1:

PR #1120, opened 2015-07-06
id:6468 comment: meta data for Debt Management Page

Guntrisoft:
- There was no translations provided, so for welsh:
  - Used the h1 title for the meta title on welsh,
  - Same english description as the welsh version

### DIFF
--- a/config/locales/debt_management/debt_management.en.yml
+++ b/config/locales/debt_management/debt_management.en.yml
@@ -1,12 +1,12 @@
 en:
-  meta_title: What to do if your debt management plan has stopped
-  meta_description: Find out what to do and where to get free help if your debt management plan has stopped due to your debt management company no longer providing a service.
-  meta_canonical: https://www.moneyadviceservice.org.uk/en/campaigns/debt-management
-  meta_alternate:
-    en-GB:
-      https://www.moneyadviceservice.org.uk/en/campaigns/debt-management
-    cy-GB:
-      https://www.moneyadviceservice.org.uk/cy/campaigns/debt-management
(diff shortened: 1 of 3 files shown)


neoeno:
Is there a reason we can't get this translated?

Guntrisoft:
I believe its mainly to concentrate SEO optimisation for the page, and getting the welsh translation wasn't deemed worth the effort.

neoeno:
Happy to get this translated for you if it's just that it's just been missed by the card-creator. I'll add it to my list if you like?

Guntrisoft:
My completeness and correctness head says yes, and the other side of me says talk to Mani/Jacqui about it :)

config/locales/debt_management/debt_management.en.yml
```diff
@@ -1,4 +1,12 @@
 en:
+  meta_title: What to do if your debt management plan has stopped
```
asmega: i'm surprised this contents of this file is not namespaced eg...
```
en:
  campaigns:
    debt_management:
      translations: go here
```
as all the locale files will get munged together into one hash which I18n lib will use

I think we're lucky this isn't causing bugs or is causing bugs that we are not aware off...

although maybe outside the scope of this ticket this should be noted and fixed

benbarnett:
:+1:

config/locales/debt_management/debt_management.en.yml
benbarnett: nice, good spot

tomas-stefano:
:+1:

PR #1124, opened 2015-07-08
id:6458 comment: Suppress breadcrumb for mobiles devices

Guntrisoft:
screen shot 2015-07-08 at 12 10 15

viewports less than 720px view port

It was felt that the breadcrumb was taking up too much screen real-estate on mobiles, and also google had an issue in their 'mobile optimised' report that the touch targets were too small on the crumb.

- ensures it only surpresses the breadcrumb visually and is still
  accessible for screen readers

### DIFF
--- a/app/assets/stylesheets/layout/common/_context_bar.scss
+++ b/app/assets/stylesheets/layout/common/_context_bar.scss
@@ -1,15 +1,16 @@
 .l-context-bar {
-  padding-top: $baseline-unit*2;
-  padding-bottom: $baseline-unit*2;
-  background-color: $breadcrumbs-background-color;
-
-  @include respond-to($mq-m) {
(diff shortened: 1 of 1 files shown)


benbarnett:
Just seen how we did this before – maybe declare as a variable? I think both make sense though... https://github.com/moneyadviceservice/publify/blob/master/app/assets/stylesheets/components/tiles/_article_tile.scss#L1

tomas-stefano:
:+1:

Guntrisoft:
After discussions with Ben, I'm going to improve this by creating some new variables in yeast which means we can nicely do respond-to(X,Y) without doing things like "minus 0.01" - which we did on the blog (publify) project too.

darokel:
43f

benbarnett:
Nice :+1:

PR #1125, opened 2015-07-09
add summer budget promo

asmega:


rnalexander:
:+1:

Guntrisoft:
The jpeg isn't optimised

Guntrisoft:
I'll send you an optimised one @asmega on slack

asmega:
what do you use to optimize it? i use `optipng` for pngs

Guntrisoft:
jpgs recently i've used http://www.jpegmini.com/

Guntrisoft:
Also for complete reference - I use http://pngcrush.com/ for PNGs and https://jakearchibald.github.io/svgomg/ for SVGs :)

asmega:
surely there must be a command line tool for jpegs instead of relying on a third party service?

Guntrisoft:
I'm sure there is, this is more for the one offs, rather than build process - we should have this built in really.

asmega:
regardless for one-offs or automation there should still be a command line tool!

but yes agreed this should be automated somehow. i'm thinking i should have global pre-commit hook

asmega:
@Guntrisoft it's now been optimised and rebased

Guntrisoft:
:+1:

PR #1126, opened 2015-07-09
TP-6413 Add debt and mental health tool to frontend

tomas-stefano:


Gemfile
```diff
@@ -49,6 +49,7 @@ gem 'car_cost_tool', '~> 1.0.4'
 gem 'contribution_calculator', '~> 0.0.1'
 gem 'cutback_calculator', '~> 0.10'
 gem 'debt_advice_locator', '~> 2.0'
+gem 'debt_and_mental_health', path: '../debt_and_mental_health'
```
tomas-stefano: I will remove the path since the PR for this repo is merged.

lib/tool_mount_point/debt_and_mental_health.rb
```diff
@@ -0,0 +1,6 @@
+module ToolMountPoint
+  class DebtAndMentalHealth < Base
+    EN_ID = 'debt-and-mental-health'
+    CY_ID = ''
```
tomas-stefano: Does not have welsh url for this tool at the moment.

Tr4pSt3R:
:+1:

Tr4pSt3R:
You may want to do something like this, ```CY_ID = 'debt-and-mental-health'``` perhaps. And a descriptive spec to say that it has no welsh alternative hence welsh locale requests will always refer to english

tomas-stefano:
I will add a descriptive comment, but I will left as blank since there is no link and if go with the welsh will link will not work as intended.

tomas-stefano:
Done.

asmega:
:+1:  May want to check if adding all the frontend assets will cause problems elsewhere

tomas-stefano:
Frontend assets? Did you mean frontend gems? The jquery rails is the same version as frontend. The only one that needs to verify it is the  jquwry mobile.

asmega:
Yeah, sorry that's what i meant

darokel:
:+1:

Tr4pSt3R:
@tomas-stefano this can now be merged if you are happy with it.

PR #1131, opened 2015-07-14
id:6504 comment: Contact Us page - add Newsletter sign-up module

Guntrisoft:
- adding option to display the newsletter signup straight away
  and not use the Newsletter.js script to place it

### DIFF
--- a/app/views/shared/_news_signup_test.html.erb
+++ b/app/views/shared/_news_signup_test.html.erb
@@ -1,4 +1,4 @@
-<div class="news-signup-test" data-dough-component="Newsletter">
+<div class="news-signup-test <% if local_assigns.has_key?(:visible) %>news-signup-test--with-image news-signup-test--in-article <% end %>" <% unless local_assigns.has_key?(:visible) %>data-dough-component="Newsletter"<% end %>>
   <div class="news-signup-test__image"></div>
   <div class="news-signup-test__container">
   <div class="news-signup-test__container__image"></div>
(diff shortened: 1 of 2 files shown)


app/views/shared/_news_signup_test.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="news-signup-test" data-dough-component="Newsletter">
+<div class="news-signup-test <% if defined?(visible) %>news-signup-test--with-image news-signup-test--in-article <% end %>" <% unless defined?(visible) %>data-dough-component="Newsletter"<% end %>>
```
tomas-stefano: Can you change from `defined?(visible)` to `local_assigns.has_key?(:visible)`?

tomas-stefano:
:+1:

app/views/shared/_news_signup_test.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="news-signup-test" data-dough-component="Newsletter">
+<div class="news-signup-test <% if local_assigns.has_key?(:visible) %>news-signup-test--with-image news-signup-test--in-article <% end %>" <% unless local_assigns.has_key?(:visible) %>data-dough-component="Newsletter"<% end %>>
```
asmega: use a ternary here

asmega: actually use a ternary and move the double quote outside of the conditional. how does that sound?

asmega: oh wait they're different attributes. my fault :+1:

asmega:
:+1:

PR #1133, opened 2015-07-14
id:6500 comment: Sign-up pages copy change

Guntrisoft:
- uses description which is stored in session (from budget planner)
  or default text


tomas-stefano:
:+1:

mconnell:
:+1:

PR #1141, opened 2015-07-16
TP-6521 add opt-in checkbox to user registration page

Tr4pSt3R:


app/views/devise/registrations/new.html.erb
```diff
@@ -108,6 +108,14 @@
             </div>
           </fieldset>
 
+          <fieldset class="form__group">
+            <div class="form__group-item">
+              <%= f.label :opt_in, class: "form__label-heading" do %>
+                <%= f.check_box :opt_in, class: "form__group-input" %>
+                <%= t("activerecord.attributes.user.opt_in") %>
```
tomas-stefano: double quotes

app/views/devise/registrations/new.html.erb
```diff
@@ -108,6 +108,14 @@
             </div>
           </fieldset>
 
+          <fieldset class="form__group">
+            <div class="form__group-item">
+              <%= f.label :opt_in, class: "form__label-heading" do %>
```
tomas-stefano: double quotes

Tr4pSt3R:
@tomas-stefano. Your recommended fixes has been made. Please take another look

Tr4pSt3R:
Hi @darokel or @asmega is it possible to have a look this please to see if that's sufficient for User-Customer mapping for CRM please? 
Thanks D

darokel:
I think you'll also need to add your fields to `/lib/core/repository/customers/cream.rb` within the `FindMapper` `call` instance method.

asmega:
@Tr4pSt3R I don't know. Can you not test it against the test box?

app/views/devise/registrations/new.html.erb
```diff
@@ -13,12 +13,12 @@
     <p><%= t("authentication.mandatory_warning") %></p>
     <%= f.validation_summary %>
 
-    <div class="registration__row">
-      <div class="registration__field">
+    <div class='registration__row'>
```
Guntrisoft: The standard is double quotes for html attributes https://github.com/moneyadviceservice/frontend-code-standards#html-markup

app/views/devise/registrations/new.html.erb
```diff
@@ -13,12 +13,12 @@
...
+      <div class='registration__field', id='first_name'>
```
asmega: no commas between attributes in HMTL

asmega: why did you add the `id`? what is it used for?

app/controllers/registrations_controller.rb
```diff
@@ -24,6 +24,7 @@ def configure_permitted_parameters
     devise_parameter_sanitizer.for(:sign_up) << :first_name
     devise_parameter_sanitizer.for(:sign_up) << :post_code
     devise_parameter_sanitizer.for(:sign_up) << :newsletter_subscription
+    devise_parameter_sanitizer.for(:sign_up) << :opt_in
```
asmega: is `opt_in` the best name? what are we opting into?

app/views/devise/registrations/new.html.erb
Tr4pSt3R: @asmega the idea was added before I realised that there was a gem called SitePrism which, was capable of picking up this element without having to explicitly set an id. 

I will remove this as well as fix the comma in the HTML attributes

asmega:
:+1:

darokel:
:+1:

PR #1144, opened 2015-07-20
Accessibility updates suggested by DAC

jonalport:
This PR includes some accessibility enhancements for the category pages, based upon DAC recommendation which can be seen below.

--
DAC Response - I would say that adding the function to the heading is a really good idea. However, a keyboard only user may become disorientated by the ‘+’ not receiving any visible link focus on tab. To me, the focus being only on the ‘Running a bank account’ and not on the ‘+’ sign (see screen shot) suggests that I will not be viewing more content. I would definitely underline the ‘+’ sign and possibly make it green so that they look related and one link.
Also, a ‘view more’ link usually directs users to another page/window. The ‘+’ sign is the standard icon for expandable content.

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_category_detail.scss
+++ b/app/assets/stylesheets/components/page_specific/_category_detail.scss
@@ -143,7 +143,12 @@
       &:focus,
       &:hover {
         outline: none;
-        text-decoration: underline;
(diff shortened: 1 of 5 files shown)


benbarnett:
Nice :+1:

asmega:
:+1: May want some Welsh in there though

PR #1147, opened 2015-07-20
id:6499 comment:fix for button on small viewport

davidtrussler:
- welsh language version of button too wide so allowed text to wrap only in this instance.
- updates previous fix for this issue which applied to button--primary, which is now replaced by button--newsletter


Guntrisoft:
:+1:

asmega:
:+1:

PR #1153, opened 2015-07-28
refactor interactors with inheritance

asmega:
- minor api change with exposing of id to ensure compatability

cc @tomas-stefano

### DIFF
--- a/lib/core.rb
+++ b/lib/core.rb
@@ -33,6 +33,7 @@ module Core
     autoload :Technical, 'core/entity/feedback/technical'
   end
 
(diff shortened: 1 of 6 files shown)


tomas-stefano:
Does NewsReader require modification too?

asmega:
it's rather different, alo looks like it's still using the old public website...

tomas-stefano:
Yes, I am seeing this now. Yes, everything good to go. Good work mate :+1: :+1: :shipit:

Tr4pSt3R:
nice and succinct :+1:

PR #1157, opened 2015-07-31
TP-6556: add contact_number to CRM

Tr4pSt3R:
depends on https://github.com/moneyadviceservice/cream/pull/13

### DIFF
--- a/lib/core/entity/customer.rb
+++ b/lib/core/entity/customer.rb
@@ -1,7 +1,7 @@
 module Core
   class Customer < Entity
     attr_accessor :first_name, :last_name, :email, :post_code, :age_range,
-                  :gender, :state, :topics, :newsletter_subscription,
+                  :gender, :state, :topics, :newsletter_subscription, :contact_number,
                   :date_of_birth, :status_code
 
     def active?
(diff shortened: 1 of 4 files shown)


tomas-stefano:
:+1:

lib/core/entity/customer.rb
```diff
@@ -33,7 +34,8 @@ def to_crm_hash
         GenderCode: { Value: GENDER_MAP[gender] },
         mas_AgeRange: { Value: AGE_RANGES_MAP[age_range] },
         BirthDate: iso8601_date_of_birth,
-        DoNotBulkEMail: !newsletter_subscription
+        DoNotBulkEMail: !newsletter_subscription,
+        Telephone2: contact_number
```
asmega: was `Telephone` used for something else?

Tr4pSt3R: Telephone2 is how it is named in CRM. That's the logical name that the CRM guys sent across. 
screen shot 2015-08-03 at 09 54 55

asmega:
:+1:

neoeno:
:+1:

PR #1166, opened 2015-08-06
[DEV APPROVED] Update rio and pencalc

neoeno:
Also sneakily made a test a little more robust (i.e. stop it failing).

Probably there's some feature flag I could disable to make it pass, but it seems like the tests should pass either way.


benbarnett:
:+1:

asmega:
:+1:

PR #1167, opened 2015-08-06
Tidying up in article newsletter form

Guntrisoft:
- there is a whole bunch of sass stuff which is no longer
required for the newsletter_signup_test (in-article) module
- They were originally there to handle different versions
  of the signup form which were A/B tested through optimizely
- We've now settled on the one which is in between article
  text, so a lot of sass is no longer required
- also removed the close button as that isn't relevant in
  this in-article version


jonalport:
Nice deleting :+1:

benbarnett:
I see red :+1:

PR #1169, opened 2015-08-07
Social sharing

davidtrussler:
6545: Social sharing icons added to article and category pages. This relies on the newly-created Dough component 'Social Sharing icons'


benbarnett:
Looks great. Is there a related PR for the new Dough component?

app/assets/stylesheets/components/common/_social-sharing.scss
```diff
@@ -0,0 +1,34 @@
+.list {
+  > li.social-sharing__item:before {
```
benbarnett: Can we move this style so it sits within the `.social-sharing` namespace? Strictly speaking, components shouldn't really be modifying other components.

app/assets/stylesheets/components/common/_social-sharing.scss
```diff
@@ -0,0 +1,34 @@
...
+    background-image: none;
+    display: none;
+  }
+}
+
+.social-sharing {
+  h2 {
+    @include body(24, 30);
+    color: $color-heading-extra-small;
+    margin-bottom: 0.5rem;
```
benbarnett: We have a thing called `$baseline-unit` that ends up being outputted as `rem`'s or `em`'s depending on whether it's the responsive or fixed stylesheet. We tend to use it for vertical rhythm, so this is a good candidate for this.

You could use it like `margin-bottom: $baseline-unit`

davidtrussler:
Yes, will be there shortly.

app/assets/stylesheets/components/common/_social-sharing.scss
```diff
@@ -0,0 +1,21 @@
+%icon {
+  .social-sharing & {
+    background-image: none;
+    display: none;
+  }
+}
+
+.social-sharing {
+  h2 {
```
benbarnett: Can we target this with a classname rather than the element directly?

app/assets/stylesheets/components/common/_social-sharing.scss
```diff
@@ -0,0 +1,21 @@
+%icon {
```
benbarnett: Is this still needed?

app/assets/stylesheets/components/common/_social-sharing.scss
```diff
@@ -0,0 +1,19 @@
+.social-sharing {
+  .social-sharing__title {
```
benbarnett: We've found that because the class names are already pretty specific, we don't actually need to nest them like this. The output is larger (although it probably gzip's out) and would just keep it inline with the rest of the CSS. Could you move the styles out so they're not nested?

benbarnett:
For reference, this depends on https://github.com/moneyadviceservice/frontend/pull/1169

benbarnett:
Looks good :+1:

tomas-stefano:
:+1: LGTM

PR #1170, opened 2015-08-10
port email anywhere validator from old site

asmega:
- dependent on https://github.com/moneyadviceservice/postcode_anywhere-email_validation/pull/3

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,7 @@ gem 'rouge'
 gem 'rubytree'
 gem 'statsd-ruby'
 gem 'uglifier', '>= 1.3.0'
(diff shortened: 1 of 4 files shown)


darokel:
I think there may be some overlap with this https://github.com/moneyadviceservice/frontend/blob/509bd65a3162cce51c108234cce97ff589d5c848/lib/validators/email.rb?

asmega:
So there is! I'll append the dns validation option on the end of that validator then

asmega:
@darokel i've now moved this to the email validator that was already present

darokel:
:shipit: :shipit:

PR #1176, opened 2015-08-14
id: 6614, comment: copy changes in the "Call us" panel

davidtrussler:


config/locales/cy.yml
```diff
@@ -153,7 +153,7 @@ cy:
           <li class="contact-panel__additional-info">Dydd Sadwrn, 9am i 1pm</li>
           <li class="contact-panel__additional-info">Dydd Sul a Gwyliau Banc, ar gau</li>
         </ul>
-      smallprint: Yr un gost â galwad arferol, os yw eich galwadau am ddim, caiff ei gynnwys.
+      smallprint: "Mae galwadau yn costio’r un faint â galw rhif 01 neu 02. Os oes gennych becyn galw “munudau am ddim”, gwiriwch gyda’ch darparwr bod rhifau 03 yn cael eu cynnwys."
```
munckymagik: Not sure if this is still the rule, but I believe the MAS preferred standard is to convert "MS-Word double-quotes" to normal ASCII quotes.

davidtrussler: I had wondered if there were issues around character encoding in YAML files and if there were standards regarding that but precedent suggested not. In this instance that character is part of the code rather than the content so shouldn't be an issue.

munckymagik: Sure, it'll probably be inconsistent because it's really easy to copy and paste these characters from Word. In previous PRs we fixed whole YAML files so they consistently used just `'` and `"` **within** the translated content. @darokel are we still trying to do this these days?

benbarnett: Was there a conclusion to this?

davidtrussler: I believe the addition is consistent with what is already there but if there is a standard I'm happy to update to match that.

benbarnett: I have actually seen these quotes used before, so as long as they're valid HTML in our charset (which I think they are?) then this seems cool.

neoeno:
:+1:

munckymagik:
:+1: sorry if I caused this to stall.

PR #1181, opened 2015-08-19
Fix wrapping commas on header links

jonalport:
Tried to do this with pseudo-elements but after much difficulty decided to stick to basics.

### DIFF
--- a/app/views/categories/_top_links.html.erb
+++ b/app/views/categories/_top_links.html.erb
@@ -2,7 +2,7 @@
 <ul class="category-link-list">
   <% category.links.each do |link| %>
     <li class="category-link-list__item">
-      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %></a>
+      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %><%= ',' unless link.equal? category.links.last %></a>
     </li>
   <% end %>
 </ul>
--- a/app/assets/stylesheets/components/page_specific/_category_link_list.scss
+++ b/app/assets/stylesheets/components/page_specific/_category_link_list.scss
@@ -30,14 +30,6 @@
   }
 }
 
-.category-link-list__item:after {
-  content: ", ";
-}
-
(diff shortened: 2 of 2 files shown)


app/views/categories/_top_links.html.erb
```diff
@@ -2,7 +2,7 @@
 <ul class="category-link-list">
   <% category.links.each do |link| %>
     <li class="category-link-list__item">
-      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %></a>
+      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %><%= ',' unless link === category.links.last %></a>
```
asmega: 2 things. should the comma have a space after it like the css? and is `==` not sufficient as the equality check?

jonalport: As far as the comma after the space, I intentionally left that out as the spacing is added more explicitly via CSS margin.

What I was trying to do with the === was an object comparison (to see if they were the same object), as opposed to a value comparison. It seems like that is not really a thing, however == does not do it either. So I have settled on ".equals?" which works. Sound good?

asmega:
:+1:

app/views/categories/_top_links.html.erb
```diff
@@ -2,7 +2,7 @@
 <ul class="category-link-list">
   <% category.links.each do |link| %>
     <li class="category-link-list__item">
-      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %></a>
+      <a class="category-link-list__item__link" href="<%= link['url'] %>"><%= link['text'] %><%= ',' unless link.equal? category.links.last %></a>
```
benbarnett: Surprised there's not a more succinct way to do that in ruby. Doesn't seem there is though!

benbarnett:
:+1:

app/views/categories/_top_links.html.erb
jonalport: Agreed!

PR #1183, opened 2015-08-20
Add IE edge meta

jonalport:
Is this the right file to add this into? How do 3rd parties pull in the head?


benbarnett:
This gets renders on the empty layout, which uses the same layout bar a few minor changes: https://www.moneyadviceservice.org.uk/en/empty

benbarnett:
The 3rd parties then pull in that page via curl or similar

benbarnett:
Also :+1:

PR #1186, opened 2015-08-21
[DEV APPROVED] consume cms redirects

asmega:
- this feature is hidden behind a feature toggle
- i have removed the blacklisted categories constraint as it seems to have no affect in current production


tomas-stefano:
Is this cover categories, corporate requests too?

tomas-stefano:
The code LGTM. Nice done

asmega:
@tomas-stefano good point. I made the change to `BaseContentReader` which affects action plans, article and videos. So it would not affect categories or corporate

asmega:
Actually it does affects corporate as that inherits from articles

tomas-stefano:
Just for curiosity, what happens if the user hit the following url '/en/tools/some-inexistent-tool'? They will request a redirect?

asmega:
- assuming no route matches it will hit the last catchall route.
- this goes here `app/controllers/catchall_controller.rb`
- checks for a redirect. if one exists use that
- otherwise raise routing error which will go through all stuff that is currently implemented (i've made no changes there)

tomas-stefano:
But the tools has a rule: https://github.com/moneyadviceservice/frontend/blob/master/config/routes.rb#L150

Will by pass that?
I know that this is not the acceptance criteria, I am just trying to see if we implment the redirects for the tools, how will be.

tomas-stefano:
Btw, LGTM this PR :+1:

darokel:
What's CMSRaw?

asmega:
@darokel the other cms repository makes certain assumptions so is rather constraining. so i've removed them and this was the only other name i could think of. Compare the 2 different files and you'll see the difference

asmega:
@tomas-stefano that will continue to go to the 501. there are also other routes that go to the 501 not implemented path. If i feature toggle all of those routes do you think that would be better?

tomas-stefano:
Can we sit down together and see all the 501?

2015-08-24 9:47 GMT+01:00 Phil Lee <notifications@github.com>:

> @tomas-stefano <https://github.com/tomas-stefano> that will continue to
> go to the 501. there are also other routes that go to the 501 not
> implemented path. If i feature toggle all of those routes do you think that
> would be better?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/moneyadviceservice/frontend/pull/1186#issuecomment-134098923>
> .
>



-- 
Atenciosamente,
Tomás D'Stefano

asmega:
@darokel I've renamed cms raw to cms api

darokel:
:+1:

asmega:
@tomas-stefano I've tidied the 501 routes in confit/routes. How does that look?

tomas-stefano:
I will take a look

config/routes.rb
```diff
@@ -278,6 +262,9 @@ def not_implemented
     end
   end
 
-  match '*path', via: :all, to: not_implemented
-
+  if Feature.active?(:redirects)
+    match '*path', via: :all, to: 'catchall#not_found'
```
Tr4pSt3R: I think this will break the ability to find old tools on the public website. 
Secondly, I'm not sure about the action name ```not_found```.

asmega: I've re-added this line https://github.com/moneyadviceservice/frontend/pull/1186/files#diff-21497849d8f00507c9c8dcaf6288b136R265 does that resolve your issue for old tools?

asmega: What would you prefer as an alternate name?

Tr4pSt3R: i will think of a better name...

Tr4pSt3R: I think you can just use the action name ```not_implemented```

asmega: is that not pretty much the same?

Tr4pSt3R: It is, but for the sake of legacy code readability. Additionally, there is not 404 ```not_found``` template on frontend. Not unless you will be rendering 404's through CMS

asmega: I've made the change

Tr4pSt3R:
@asmega I will have a go at testing it tomorrow just to make sure that it will still allow work with the old tools

asmega:
@Tr4pSt3R how is your testing going?

asmega:
@Tr4pSt3R are you happy with your testing?

app/controllers/catchall_controller.rb
```diff
@@ -0,0 +1,17 @@
+class CatchallController < ApplicationController
+  def not_implemented
+    interactor.call do |error|
+      if error.redirect?
+        return redirect_to error.location, status: error.status
+      end
+    end
+
+    raise ActionController::RoutingError.new('Not Found')
```
Tr4pSt3R: In ```raise ActionController::RoutingError.new('Not Found')```, is the text ```"Not found" ``` the right error message?

asmega: this uses the current implementation found here https://github.com/moneyadviceservice/frontend/blob/master/app/controllers/application_controller.rb#L33 that rest of the site uses

Tr4pSt3R: Thanks @asmega, just checking.

Tr4pSt3R:
:+1:

PR #1194, opened 2015-09-04
fix 404 catchall when cms content not found

asmega:
the cms api throws an error when it doesn't recognize certain urls
for now we handle this to return a 404

the better fix is to make the cms not have this constraint
and return a 404 when the resoure cannot be located
however this needs further looking into in case there are compatability
issues


darokel:
:+1:

tomas-stefano:
:+1:

PR #1196, opened 2015-09-07
add timings to the build script

asmega:
- my limited bash knowledge to add timings to steps of the build process to identify bottlenecks

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,11 +24,11 @@ rm -rf public/assets vendor/cache coverage log/* tmp/*
 
 echo "Running Bundle package"
 echo "----"
-bundle package --all
+time bundle package --all
 
 echo "Running Bower cache clean"
 echo "----"
-bower cache clean
+time bower cache clean
 
 echo "Purging bower components"
 echo "----"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -18,38 +18,86 @@ cat > public/version <<EOT
...
+echo "================="
+
+START_TIME=$SECONDS
 echo "Purging bower components"
 echo "----"
 [ -d vendor/assets/bower_components ] && rm -r vendor/assets/bower_components
 
+ELAPSED_TIME=$(($SECONDS - $START_TIME))
+echo "================="
+echo "Purging bower componments took $ELAPSED_TIME seconds"
```
Tr4pSt3R: typo *components

build.sh
```diff
@@ -18,38 +18,86 @@ cat > public/version <<EOT
...
 cd ..
 /usr/local/rpm_builder/create-rails-rpm $artifact_name $artifact_name $version
 
+ELAPSED_TIME=$(($SECONDS - $START_TIME))
+echo "================="
+echo "RPM took $ELAPSED_TIME seconds"
+echo "================="
+
 #prune packaged gems
 rm -rf vendor/cache
```
tomas-stefano: `time /usr/local/rpm_builder/create-rails-rpm $artifact_name $artifact_name $version` will work right?

build.sh
asmega: fixed

Tr4pSt3R:
:+1: :100:

build.sh
asmega: yeah that's a better idea

asmega: how about now? i only timed the longer running commands

tomas-stefano:
:+1: :ship:

PR #1199, opened 2015-09-09
[DEV APPROVED]use in-memory sqlite database in test mode

asmega:
the time is roughly the same on my local box although I've only done a few runs. however in theory it should be quicker as there is no disk io overhead and assuming the test box has enough memory.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -8,7 +8,7 @@ development:
 
 test:
   adapter: sqlite3
-  database: db/test.sqlite3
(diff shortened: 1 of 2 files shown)


munckymagik:
@asmega what's driving the change to SQLite for the tests? The PR looks fine of course, but you say the performance is about the same locally, so is there another problem that's solving?

munckymagik:
:+1:

darokel:
:+1:

PR #1204, opened 2015-09-09
remove old public website articles repository

asmega:


benbarnett:
Can you explain a bit more about what these PR's are doing for us laypeople?

asmega:
These are all the related repositories/drivers to get content from the old public website. Since most of these have been swapped over for the new cms, these drivers are no longer used therefore not needed.

benbarnett:
Yep thanks! :+1:

PR #1208, opened 2015-09-16
remove redundant assets in build script

asmega:


benbarnett:
Was this breaking something or just a spring clean?

:+1:

Tr4pSt3R:
:+1: :leaves:  
I would say Autumn cleaning.

asmega:
i'm trying to get rid of assets we don't need when we precompile the assets. not sure if there's an option i can pass to the precompile to ignore these but the end result will be the same.

so no need to precompile it and it wont be needed in the built package. so a slight optimisation.

benbarnett:
Really we should be cleaning up the bower.json's to use `dependencies` and `devDependencies`, so the latter won't be pulled in when run in production mode. This is a simpler way of achieving this but probably won't help the performance as much?

darokel:
Agreed with @benbarnett. That should be the first thing to do

asmega:
@benbarnett i tried that however did not seem to work. i'm guessing that splitting those dependencies is only used by bower if frontend manifest was being used in another project.

i'm happy for you to prove me wrong and get a proof of concept to show me it working.

tomas-stefano:
What is the status of this PR? :)

asmega:
hung. happy to close as there doesn't appear to be a better solution

tomas-stefano:
'Close' did you mean merge? For me is fine but I don't know the others. Wdyt @benbarnett and @darokel? Merge or discuss more the subject?

asmega:
Close as there is some dissonance in the crowd

Tr4pSt3R:
I say go ahead and merge this

benbarnett:
@asmega yes you're right - there are many bower configs and it only needs one of them to have these things still in `dependencies` to end up on the build. Happy to be pragmatic and just merge this, but should maybe extend the comment to say how this is a failsafe and really should be fixed at source

munckymagik:
"Should we close or should we merge? Duh duh Duh duh, Duh duh Duh!" -- A song by the (Git) Stash.

PR #1209, opened 2015-09-16
Install cost_calculator_builder engine from gem and mount

tom-kitchin:
Mount Cost Calculator Builder


db/schema.rb
```diff
@@ -425,6 +498,19 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "friendly_id_slugs", force: true do |t|
```
rnalexander: All of the other new table definitions are name-spacing themselves to the cost_calculator_builder, but not this one, any reason why?

tom-kitchin: I think that's because it's a generated table from the friendly_id module which provides slugs instead of IDs in paths. Good catch, I'll see if I can make it namespace properly.

tom-kitchin:
Rebased after change due to migration changes.

rnalexander:
Nice Job, looks very clean!  :+1:

darokel:
Could we add the feature flag to this file? https://github.com/moneyadviceservice/frontend/blob/master/config/features.yml.sample It's just an index so other devs can use it as a reference of what to enable for what functionality/tool

tom-kitchin:
Already have done :)

PR #1219, opened 2015-09-25
Fix opt in db migration

Tr4pSt3R:
Fixes the issue we experienced and the horrors of setting defaults at the database level. This fix maintains the opt-in for market research field being set to true automatically. This assertion is on the form only. It's done neither through callbacks nor database defaults.


app/views/devise/registrations/new.html.erb
```diff
@@ -133,7 +133,7 @@
         <fieldset class="form__group">
           <div class="form__group-item">
             <%= f.label :opt_in_for_research, class: "form__label-heading" do %>
-              <%= f.check_box :opt_in_for_research, class: "form__group-input" %>
+              <%= f.check_box :opt_in_for_research, {checked: true}, class: "form__group-input" %>
```
tomas-stefano: I would add a space between the hashes. `{ checked: true }`

tomas-stefano: What will be done with the edit form?

Tr4pSt3R:
Closed. Messy schema file. New PR here: https://github.com/moneyadviceservice/frontend/pull/1225

PR #1224, opened 2015-09-28
allow news articles to be redirected

asmega:


spec/controllers/news_controller_spec.rb
```diff
@@ -30,15 +30,34 @@
...
       it 'raises an ActionController RoutingError' do
-        allow(Core::NewsArticleReader).to receive(:new) { ->(&block) { block.call } }
-
         expect { get :show, id: 'foo', locale: I18n.locale }.to raise_error(ActionController::RoutingError)
       end
     end
+
+    context 'when news article is redirected' do
+      before do
+        allow_any_instance_of(Core::NewsArticleReader).to receive(:call).and_yield(OpenStruct.new(status: 301, location: 'https://example.com', redirect?: true))
```
Tr4pSt3R: Give this OpenStruct object a name perhaps? 

```
redirect_path = OpenStruct.new(status: 301, location: 'https://example.com', redirect?: true)
allow_any_instance_of(Core::NewsArticleReader).to receive(:call).and_yield(redirect_path)
```

munckymagik: Just to note, and I'm not suggesting you necessarily need to action this as part of this PR, but some of these lines are a bit long, longer than the max width of GitHub's PR diff. As GitHub's diffs don't line-wrap we're having to horizontally scroll to review the entire line.

So yes, intermediate variables or even simply breaking chained statements onto multiple lines would be a big help for review.

munckymagik:
:+1:

spec/controllers/news_controller_spec.rb
asmega: I've now extracted these to variables which has shortened the lines a bit

Tr4pSt3R:
:+1:

PR #1225, opened 2015-09-28
Fix opt in db migration

Tr4pSt3R:
Fixes the issue we experienced and the horrors of setting defaults at the
    database level. This fix maintains the opt-in for market research field being
    set to true automatically. This assertion is on the form only. It's done neither
    through callbacks nor database defaults.


tomas-stefano:
:+1:

munckymagik:
:+1:

Tr4pSt3R:
Awesome :+1: 
We can finally put this ethics debate to bed

PR #1227, opened 2015-09-29
limit static pages to contact us

asmega:
- this removes code that is no longer used as we only have 2 static pages. these are `contact-us` and the welsh equivalent
- this also solves the problem of static redirects handled by the cms. as these routes will no longer be routed by the application the catchall route and controller will take care of these redirects.


tomas-stefano:
:+1:

munckymagik:
:+1:

PR #1230, opened 2015-09-30
allow categories to be redirected with the cms

asmega:
- also added missing tests to ensure coverage


rnalexander:
:+1: :shipit:

PR #1231, opened 2015-10-01
catchall controller passes format to cms

asmega:
for example this will allow urls with the extension .pdf
to be redirected


rnalexander:
:+1: :shipit:

tomas-stefano:
:+1:

PR #1232, opened 2015-10-01
remove pointless route

asmega:
this route is pointless because it will eventually be caught by a bigger net anyway. 
```
if Feature.active?(:redirects)
    match '*path', via: :all, to: 'catchall#not_implemented'
  else
    match '*path', via: :all, to: not_implemented
  end
```

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -269,8 +269,6 @@ Rails.application.routes.draw do
     end
   end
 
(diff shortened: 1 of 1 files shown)


Tr4pSt3R:
:+1: :100:

rnalexander:
:+1: :shipit:

PR #1239, opened 2015-10-06
Adds default iframe height for non-javascript users

benbarnett:
Will be overridden by the iframe resizer javascript code


jonalport:
Lovely work :+1:

johnplayer1982:
:+1:

PR #1240, opened 2015-10-06
Adding optimizely embed code to empty page

benbarnett:
An agency for Optimizely are running a training session on Friday and need an empty page on production to demonstrate some tests with. We have decided to use the https://www.moneyadviceservice.org.uk/en/empty template for this as it is only used by third parties in a non-automated fashion.

This code simply drops in the embed code to those pages using the `hide_elements_irrelevant_for_third_parties` flag, which is only set to true on the `empty` controller.

### DIFF
--- a/app/views/layouts/_base.html.erb
+++ b/app/views/layouts/_base.html.erb
@@ -10,6 +10,12 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <%= optimizely_include_tag if Rails.env.production? && Feature.active?(:optimizely) %>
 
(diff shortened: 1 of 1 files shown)


johnplayer1982:
Nice :+1:

jonalport:
Variable naming like a boss :+1:

benbarnett:
Virtuous verbosity

jonalport:
alluring alliteration

jonalport:
nice willy

darokel:
:joy:

PR #1243, opened 2015-10-07
Mount quiz

Tr4pSt3R:


pmcostello:
:+1: looks good

Gemfile
```diff
@@ -59,9 +59,9 @@ gem 'feedback', '~> 0.3.1'
 gem 'mortgage_calculator', '~> 1.3.5'
 gem 'payday_loans_intervention', '~> 1.4'
 gem 'pensions_calculator', '~> 1.0.0'
+gem 'quiz', git: 'git@github.com:moneyadviceservice/quiz.git'
```
benbarnett: Any reason we're using git rather than the gemserver?

Tr4pSt3R: @benbarnett there's another gem in RubyGems so to avoid name clash we used this. We could just as well have defined the source

benbarnett: Ah ok, cool

benbarnett:
:+1:

PR #1252, opened 2015-10-13
[DEV APPROVED]Updating engine versions for multiple tasks

johnplayer1982:
Updating versions for the following (engine pipelines triggered and passed):
RIO - Removing BETA
https://github.com/moneyadviceservice/rio/pull/78

savings_calculator - Remove feedback link for partners
https://github.com/moneyadviceservice/savings_calculator/pull/22

timelines - Add meta descriptions & separate canonical for tool
https://github.com/moneyadviceservice/timelines/pull/61

Not 100% sure why autoprefixer-rails (updates on bundle update savings_calculator) and pensions_calculator-calculations (updates on bundle update rio) are updating also? I will roll these back before merge if necessary? @benbarnett

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -43,7 +43,7 @@ GIT
 
 GIT
   remote: git@github.com:moneyadviceservice/rio.git
-  revision: f00200848c21897f8eb8f8ea4af18cf1a100d555
+  revision: 22ce2aa4d45365f8d8eb91b6b87fe774c55d1869
   specs:
     rio (0.0.3)
       bowndler
@@ -53,7 +53,7 @@ GIT
 
 GIT
   remote: git@github.com:moneyadviceservice/timelines.git
-  revision: f74ec52038d738a5b7f1f56e78e090c0080e335f
+  revision: 7b5f7d89703d7904b05f4ca24ec15a8f945be0de
   specs:
     timelines (1.2.2)
       bowndler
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -137,7 +137,7 @@ GEM
     ast (2.0.0)
     astrolabe (1.3.0)
       parser (>= 2.2.0.pre.3, < 3.0)
-    autoprefixer-rails (5.2.0)
+    autoprefixer-rails (6.0.3)
```
rnalexander: autoprefixer isn't one of our packages and it's a full version upgrade, is this needed by one of the other tools?

johnplayer1982: I bundle updated each of the tools and watched for changes, autoprefixer version updated with savings_calculator

rnalexander: Gaah.  Missed the note above, apologies.  I'm going to try this locally and if it doesn't barf I'll thumb it and put it in the queue.

rnalexander:
OK, I've pulled this into master on my local machine and built it and ran both rspec and cucumber tests and they all came up green.  So I'm happy with putting this through.

I know this is all in the same file but please separate these upgrades in the future.  It'll make testing and approval even easier.  (It also means that if something *does* go wrong with one of the upgrades we won't be forced to pull back ALL of them to fix it.)

:+1:

PR #1255, opened 2015-10-13
6457 newsletter in article

nmiddleweek:
This is a rework of the inpage article newsletter sign up component for mobile/ desktop.

It's not being displayed under Welsh pages.


app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+    border-left: 0;
+    padding-left: 0;
+  }
+
+  .news-signup-inpage__header {
+    @include clearfix;
+    display: block;
+    position: relative;
+
+    .news-signup-inpage__text-container {
```
benbarnett: Do these blocks need to be nested within the header?

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+      position: relative;
+      margin-left: 0;
+
+      @include respond-to($mq-m) {
+        margin-left: 140px;
+      }
+    }
+
+    .news-signup-inpage__title {
```
benbarnett: Same here re nesting

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+        @include body(18, 24);
+        margin: 20px 20px;
+      }
+
+      @include respond-to($mq-xl) {
+        @include body(20, 24);
+        margin: 20px 40px 15px 40px;
+      }
+
+      span {
```
benbarnett: Would be good to try an use a classname for this rather than the element directly

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+
+      @include respond-to($mq-l) {
+        width: 40%;
+        margin-right: $baseline-unit;
+        margin-bottom: 0;
+        text-align: left;
+      }
+    }
+
+    .button--newsletter {
```
benbarnett: This is where the layout vs component stuff gets a bit tricky. Strictly speaking components shouldn't be styling other components. In this situation you can use a layout class (e.g. `l-news-signup-inpage`) which will position and style elements like this one does. We sort of accepted when coming up with this concept that we would sacrifice additional wrapping divs for clarity in the CSS. Might be easier to chat through this one when you have a minute.

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+
+    @include respond-to($mq-l) {
+      margin-top: 0;
+    }
+
+    .news-signup-inpage__privacy-text {
+      @include body(13, 16);
+      margin-bottom: 0;
+
+      a::after {
```
benbarnett: Same here re styling elements directly, could use a classname for this.

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+        margin-bottom: -4px;
+      }
+
+      span {
```
benbarnett: And here, classname preferred over element

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+    padding: 20px;
+  }
+
+  .news-signup-inpage__container__intro h3 {
+    @include body(22, 26);
+    color: $color-green-dark;
+    margin-top: 0;
+  }
+
+  .news-signup-inpage__container__intro p {
```
benbarnett: I think favour the nesting approach for this to make it more readable...

```scss
.news-signup-inpage__container__intro {
  h3 {
  }

  p {
  }
}
```

If we could also use class names rather than elements here then that'd be preferable too. However if this is CMS generated content then that's totally ok to target elements. (Not sure if this is yet)

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+    margin-bottom: $baseline-unit*2;
+
+    @include respond-to($mq-news-signup-inpage-article) {
+      width: 55%;
+      display: inline;
+      margin-bottom: $baseline-unit;
+    }
+  }
+
+  .news-signup-inpage__container__form .button {
```
benbarnett: Same here, personally prefer wrapping all these statements in a single `.news-signup-inpage__container__form`, but this is just a personal preference and up to you

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+
+          @include respond-to($mq-3col-layout) {
+            min-width: 220px;
+          }
+
+          @include respond-to($mq-l) {
+            min-width: 280px;
+          }
+
+          @include respond-to($mq-xl) {
```
benbarnett: These seem very specific mediaqueries, is there any way this can be simplified? No worries if not just curious.

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+            min-width: 300px;
+          }
+
+        }
+      }
+    }
+
+    .news-signup-inpage__form-container {
+      input[type="email"] {
+        border-color: #d11d32;
```
benbarnett: These should be colour variables declared in variables like this in `_variables.scss`:

```scss
// you can generate a colour name from here: http://chir.ag/projects/name-that-color/#d11d32
$whatever-this-colour-is-called: #d11d32;

// then you use a map so we reference this colour in a meaningful way:
$news-signup-inpage-field-focus: $whatever-this-colour-is-called;
```

Then back in this file you just reference that:

```scss
input[type="email"] {
  border-color: $news-signup-inpage-field-focus;
}
```

We should consider adding this to the styleguide too if this is a new way of styling fields (this is probably the stuff you were talking about the other day?)

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+
+    .news-signup-inpage__title {
+      @include respond-to($mq-m) {
+        display: none;
+      }
+    }
+    .news-signup-inpage__supplementary-text {
+      display: none;
+    }
+    .news-signup-inpage__form-container {
```
benbarnett: Should separate statements with a line break

app/views/shared/_news_signup_inpage.html.erb
```diff
@@ -0,0 +1,70 @@
+<div class="news-signup-test
```
benbarnett: Are these referencing the wrong class names now? What's the `-test` one for?

app/views/shared/_news_signup_inpage.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    <div class="news-signup-test__container__form">
+      <%= form_tag(main_app.newsletter_subscription_path, class: 'js-news-signup-test-form', remote: true) do %>
+        <%= hidden_field_tag :authenticity_token, form_authenticity_token, id: 'test_authenticity_token' -%>
+        <%= hidden_field_tag 'subscription[ga_categoryid]', 'Newsletter SignUp Test', id: 'subscription[test_ga_categoryid]' -%>
+        <%= hidden_field_tag 'subscription[category]', 'in_article', id: 'subscription[test_category]' -%>
+        <%= label_tag('subscription[email]', 'Enter your email address', for: 'news-signup-email', class: 'newsletter-signup__container__form__label') %>
+        <%= email_field_tag('subscription[email]', nil, id: 'news-signup-email', class: 'news-signup-test__container__form__email js-news-signup-test-input') %>
+        <%= button_tag('Sign up now', type: 'submit', class: 'button button--primary') %>
+      <% end %>
+      <p class="news-signup-test__container__form__privacy">Privacy promise: We will never share your data or spam
```
benbarnett: Should this be localised?

app/views/shared/_news_signup_inpage.html.erb
```diff
@@ -0,0 +1,70 @@
...
+
+        <%= email_field_tag('subscription[email]', nil, id: 'news-signup-email', class: 'js-news-signup-test-input', placeholder: t('newsletter_subscriptions.signup_inpage.controls.email.placeholder')) %>
+        <%= button_tag(t('newsletter_subscriptions.signup_inpage.controls.button.text'), type: 'submit', class: 'button button--newsletter') %>
+      </div>
+    <% end %>
+
+    <div class="js-news-signup-inpage__footer news-signup-inpage__footer">
+      <p class="news-signup-inpage__privacy-text">
+        <span><%= t('newsletter_subscriptions.signup_inpage.privacy_text') %></span>
+        <span><%= t('newsletter_subscriptions.signup_inpage.privacy_text_2') %>
```
benbarnett: Just to make this a bit more readable can we drop both spans onto block level?

```erb
<span>
  <%= t('newsletter_subscriptions.signup_inpage.privacy_text') %>
</span>
<span>
  <%= t('newsletter_subscriptions.signup_inpage.privacy_text_2') %>
  ...
</span>
```

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+        @include body(16, 21);
+        margin: 20px 20px 10px 20px;
+      }
+
+      @include respond-to($mq-l) {
+        @include body(18, 24);
+        margin: 20px 20px;
+      }
+
+      @include respond-to($mq-xl) {
```
benbarnett: Again here these seem quite tightly coupled... totally ok if this is how it works with the design, just worth asking the question to see if these can be made a bit more generic.

Another thing to note is that for vertical rhythm, we have `$baseline-unit` to keep things even for margin-top's and bottoms etc.

benbarnett:
Could you post a screenshot or two of how this looks as well? It can be really useful when reviewing the code

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+        }
+        @include respond-to($mq-l) {
+          position: relative;
+          top: auto;
+          left: auto;
+          transform: none;
+          margin: auto;
+        }
+      }
+
```
jonalport: Remove LB here?

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+          transform: none;
+          margin: auto;
+        }
+        @include respond-to($mq-3col-layout) {
+          position: absolute;
+          top: 50%;
+          left: 50%;
+          transform: translate(-50%, -50%);
+          width: 100%;
+        }
```
jonalport: Add LB here?

app/assets/stylesheets/components/common/_news_signup_inpage.scss
```diff
@@ -0,0 +1,531 @@
...
+    @include body(22, 26);
+    color: $color-green-dark;
+    margin-top: 0;
+  }
+
+  .news-signup-inpage__container__intro p {
+    @include body(14, 14);
+  }
+
+  .news-signup-inpage__container__form input[type="email"] {
```
jonalport: Could this be simplified to something like

```
.news-signup-inpage__email-input
```

?

benbarnett: Agree but for specificity reasons you'll still need the `[type="email"]` thing.

This is because the user agent stylesheets that come with the browser style fields like this at that level, so in order to override we always need to use that type selector to beat it. \o/

nmiddleweek:
I'm going to take the afternoon off to refactor this :)

app/views/shared/_news_signup_inpage.html.erb
nmiddleweek: the -test stuff is from the older version. This may change before the WIP is removed however we are A/B Testing the old versus new so need to work out a nice way of showing one or the other.

benbarnett: Cool cool

app/views/shared/_news_signup_inpage.html.erb
nmiddleweek: This is from the previous 'component' and is already live on the site. I'm guessing it was probably not localised previously because it's English only. But I will move this text into localisation as per the above.

nmiddleweek:
I've already rebased and squished. I'm going to leave these commits in place. Cheers.

benbarnett:
Re-applying [WIP] as there is code in here that can't be merged.

benbarnett:
:+1:

davidtrussler:
:+1:

nmiddleweek:
Just rebase onto master

PR #1259, opened 2015-10-16
[WIP] Update the SVG USE partial to include all necessary classes

jonalport:
There was a class missed out when the fallback etc code was added to this partial, so the CSS wasn't finding the appropriate class and hence the pluses weren't changing to minuses on the category pages.

@nmiddleweek  - can you check that I haven't broken the fallback stuff? :)

### DIFF
--- a/app/views/shared/svg/_use_icon.html.erb
+++ b/app/views/shared/svg/_use_icon.html.erb
@@ -3,25 +3,25 @@
 <% svg_fallback_support = local_assigns[:svg_fallback_support].nil? ? false : true %>
 
 <%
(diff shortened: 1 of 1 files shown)


benbarnett:
:+1:

johnplayer1982:
Pulled branch and tested locally :+1:

benbarnett:
Will wait for @nmiddleweek to verify this is ok with him before merging.

nmiddleweek:
We need to review this and it's BEM notation, it needs revising slightly and then the related styles updating.

PR #1267, opened 2015-10-21
port 503 error page from public website

asmega:


benbarnett:
Can we just serve up the new 500 page we made a month or two ago?

asmega:
@benbarnett no, that page is dynamic. this error page is for when the application server is down so needs to be completely static

benbarnett:
Agree but I think there is a static one somewhere. Bear with me.

benbarnett:
Seems it's no longer with us.

:+1:

Tr4pSt3R:
:100: :+1:

PR #1268, opened 2015-10-21
Fix for SVG icons not rendering in Safari

nmiddleweek:
Move the SVG sprites to underneath the top BODY tag but wrapped IE conditionals so it always renders unless IE < 9.


benbarnett:
:+1:

asmega:
:+1:

PR #1269, opened 2015-10-21
[DEV APPROVED] Updating IE8 hero image css

johnplayer1982:
Updating hero slot bg image to use conditional @if $responsive == false { … } rather than
.ie8 & { … }


benbarnett:
:+1:

asmega:
is this file/partial used to generate multiple css files?

johnplayer1982:
I don't think so, @benbarnett can you confirm?

benbarnett:
Nope just for this component, should be fine

nmiddleweek:
:+1:

PR #1271, opened 2015-10-26
Fix builds with newer Bundler.

misaka:
Newer versions of bundler create .bundle/config which is taking priority
over env var settings, BUNDLE_WITHOUT in particular. This breaks builds
when test.sh bundles without the "test" group and that state gets
saved. The fix, until Bundle 2 is released, is to rm the config file:
https://github.com/bundler/bundler/issues/3288

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,7 +20,7 @@ EOT
 
 echo "Cleaning temporary files"
 echo "----"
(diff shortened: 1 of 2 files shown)


darokel:
:+1:

pmcostello:
:+1:

PR #1272, opened 2015-10-26
remove mastalk dependency

asmega:
this is no longer needed as frontend acts a dumb client that merely
consumes html that is passed to it

the server (cms) does all the processing needed to generate the correct
html

previously the client (frontend) would get markdown which is would
process to HTML

this change removes the double processing of data


tomas-stefano:
Nice! :+1: :+1:

benbarnett:
Nice one :+1:

PR #1280, opened 2015-11-05
Replace Rackspace CDN with Azure asset hosting

misaka:
Tried to find a way to refactor this nicely but it didn't seem worth the effort in the end. The URLs in the localised translations would be a particular pain, having to supply additional parameters to `t()` calls, if anyone can point me to an easy way of parameterising the URLs there I may update the PR.


rnalexander:
:+1:

tomas-stefano:
Is just this few on PR or have more? If just these few, looks good to me :+1:

misaka:
@tomas-stefano I put through a couple other PRs to change `config.asset_host`, those have already been merged. This was just a mop-up of the rest of the statically-set Rackspace URLs.

tomas-stefano:
@misaka Ok then. Thanks for the explanation.

PR #1283, opened 2015-11-09
Lock devise and update advice plans

rnalexander:
Also contains the mysql2 pin as it's finally reached it was all the way up.  Rspec and Cucumber tests pass.


tomas-stefano:
:+1:

davidtrussler:
:+1:

PR #1285, opened 2015-11-10
Bugsnag

darokel:
We should really have error reporting for this project. Now that we've paid for Bugsnag, it's about time we configure it for the responsive site. 

The API key is added to Puppet. https://github.com/moneyadviceservice/scripts/pull/449

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'rails', '4.1.8'
 
 gem 'activerecord-session_store'
 gem 'aes'
+gem 'bugsnag'
 gem 'cream', '~> 0.1.0'
 gem 'delayed_job_active_record'
 gem 'devise', '3.2.4'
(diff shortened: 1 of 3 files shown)


paulanthonywilson:
:+1:

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'rails', '4.1.8'
 
 gem 'activerecord-session_store'
 gem 'aes'
+gem "bugsnag"
```
munckymagik: Are we ok to use double-quotes here in the Gemfile these days? If so ignore me.

Otherwise I'd like to whisper the following:

"quotes" (whispered).

Tr4pSt3R:
:+1:

tomas-stefano:
:+1:

Gemfile
darokel: Good catch mate, will update.

tomas-stefano: Yes. we use single quotes. After this change the PR is good to merge.

asmega:
:+1:

PR #1289, opened 2015-11-11
Remove redirects feature flag

tomas-stefano:
Since the redirects feature is proving it is working nicely,
I removed the feature flag since we don't need to use that anymore


darokel:
:+1:

Tr4pSt3R:
:100: :+1:

PR #1290, opened 2015-11-11
Fix annuities request spec

tomas-stefano:
* The spec was broken only when the feature flag was turn off.
Fixing the matcher to catch the not found exception solves the problem


Tr4pSt3R:
Round 3: 
:+1:

munckymagik:
:+1:

tomas-stefano:
mk

PR #1291, opened 2015-11-11
fix link to car cost tool in campaign page

asmega:
- unfortunately no test, as I can't figure it out. the named route seems to be missing even with the feature toggle enabled
- also i have no idea why specific campaign knowledge has filtered through to this generic decorator


Tr4pSt3R:
:+1:

tomas-stefano:
:+1:

PR #1294, opened 2015-11-16
Change the matcher verification for annuites spec

tomas-stefano:
* The spec was renamed to requests so the route to does not exist
here. Changing the matcher will fix the spec.
* This only spec can be runned with anuites feature flag turn on


Tr4pSt3R:
:+1:

asmega:
:+1: fyi ``to be_success`

tomas-stefano:
@asmega done. Can you review again, please?

asmega:
:+1:

Tr4pSt3R:
:+1:

PR #1299, opened 2015-11-18
[DEV APPROVED] 6779 fix newsletter icon

nmiddleweek:
Use the partial instead to support SVG and fallback...

`<%= render 'shared/svg/use_icon',
               icon: 'email',
               svg_fallback_support: true
    %>
`

There is no double-density sprite at the moment so I had to force single sized on double density screens using background-size: auto.


rnalexander:
:+1:

johnplayer1982:
:+1:

PR #1308, opened 2015-11-30
id: 6797 homepage columns border browser zoom bug

slaywell:
Using CSS pseudo-elements rather than <div>s to show the borders on the three-column footer on the homepage.

### DIFF
--- a/app/assets/stylesheets/layout/common/_contact_panels.scss
+++ b/app/assets/stylesheets/layout/common/_contact_panels.scss
@@ -18,7 +18,6 @@
       content: "";
       position: absolute;
       border-right: 1px solid $color-newsletter-input-border;
-      height: 0;
       height: 400px;
       left: 33.3333%;
     }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/layout/common/_contact_panels.scss
```diff
@@ -11,6 +12,21 @@
...
   border-top: 0;
   margin-top: 0;
+
+  @include respond-to($mq-l) {
+    &:before, &:after {
+      content: "";
+      position: absolute;
+      border-right: 1px solid $color-newsletter-input-border;
+      height: 0;
+      height: 400px;
```
davidtrussler: Should there be two values for 'height', or should the 'height:0' declaration be outside the media queries?

johnplayer1982:
Pulled branch and tested locally, IE FF chrome & safari :+1:

app/assets/stylesheets/layout/common/_contact_panels.scss
slaywell: Good spot, I've removed this line now. It's unnecessary, since the before / after elements are only shown within the media query anyway.

davidtrussler:
:+1:

PR #1311, opened 2015-12-02
Removing flickering cucumber test

slaywell:
The ‘Display sticky newsletter’ feature passes locally but is failing on the build server, reason being a JavaScript error linked to PhantomJs. After trying to specify an exact version of PhantomJs we are still not able to resolve the error.

Removing this test as per discussion with @rnalexander

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "karma-spec-reporter": "0.0.13",
     "karma-osx-reporter": "*",
     "mocha": "~1.15.1",
-    "phantomjs": "1.9.7-1",
+    "phantomjs": ">= 1.9.7-1",
     "sinon": "^1.9.0"
   },
     "scripts": {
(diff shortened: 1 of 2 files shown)


darokel:
I don't think we should remove feature specs if they're still needed. A better way to do it would be to mark this spec as pending.

darokel:
Use `@wip` in the scenario

slaywell:
Makes sense, done. @darokel @rnalexander

package.json
```diff
@@ -18,7 +18,7 @@
     "karma-spec-reporter": "0.0.13",
     "karma-osx-reporter": "*",
     "mocha": "~1.15.1",
-    "phantomjs": "1.9.7-1",
```
Tr4pSt3R: good stuff. Should ensure that we never get caught again with phantomjs upgrades

Tr4pSt3R:
:+1:

package.json
slaywell: Unfortunately not, this is back to how it was before. We tried to resolve it last night by specifying an exact version of PhantomJs but we still couldn't get the test to run. That's why we're removing this test (setting to WIP)!

tomas-stefano:
:+1:

PR #1313, opened 2015-12-04
[DEV APPROVED] Removes sticky newsletter from selected pages

Tr4pSt3R:
* removes sticky footer newsletter from home page
  * removes sticky footer newsletter from categories page

[Review on Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1313)

### DIFF
--- a/app/helpers/sticky_newsletter_visibility.rb
+++ b/app/helpers/sticky_newsletter_visibility.rb
@@ -1,22 +1,111 @@
 module StickyNewsletterVisibility
   class StickyNewsletterFilter
-
-    BLACKLIST_FILTERED = [ '',
-                           'categories',
-                           'corporate',
-                           'tools',
-                           'videos',
-                           'corporate_categories',
-                           'sitemap',
-                           'users'
+    BLACKLIST_FILTERED = ['',
+                          'categories',
+                          'corporate',
+                          'tools',
+                          'videos',
+                          'corporate_categories',
+                          'sitemap',
+                          'users'
                          ]
+
+    BLACKLIST_ARTICLES = %w(
+      making-or-reviewing-your-will-after-someone-dies
+      coping-with-debt-after-your-partner-dies
+      claiming-bereavement-and-other-benefits-after-someone-dies
+      coping-with-a-drop-in-household-income-after-your-partner-dies
+      dealing-with-the-debts-of-someone-who-has-died
+      when-should-you-use-a-probate-specialist
+      calculating-and-paying-tax-after-someone-dies
+      what-to-do-about-someones-pension-when-theyve-died
+      how-to-work-out-the-value-of-someones-estate
+      paying-for-a-funeral
+      funeral-plans
+      how-much-does-a-funeral-cost
+      pension-and-retirement-planning-following-a-bereavement
+      coping-with-grief-when-someone-dies
+      reviewing-your-insurance-after-a-bereavement
+      what-to-do-when-someone-dies-a-checklist
+      reviewing-your-savings-and-investments-after-a-bereavement
+      making-the-most-of-your-inheritance
+      taking-over-the-budget-after-your-partner-dies
+      sorting-out-someones-estate-where-to-start
+      help-manage-the-money-of-someone-youre-caring-for
+      support-services-available-to-carers
+      financial-support-for-young-carers
+      benefits-and-tax-credits-you-can-claim-as-a-carer
+      when-does-someone-need-formal-help-with-managing-their-money
+      help-someone-informally-with-budgeting-and-day-to-day-money
+      if-the-person-you-want-to-help-can-still-manage-their-money
+      if-the-person-you-want-to-help-has-lost-mental-capacity
+      how-to-sort-out-your-money-if-you-become-ill-or-disabled
+      support-to-help-you-keep-your-job-when-ill-or-disabled
+      help-to-find-work-if-youre-disabled
+      support-to-help-you-study-when-ill-or-disabled
+      disability-and-sickness-benefits-check-your-entitlements
+      personal-independence-payment-an-introduction
+      universal-credit-for-disabled-people
+      financial-support-if-you-or-your-child-has-a-disability
+      how-to-appeal-against-a-disability-benefits-decision
+      funding-to-adapt-your-home-for-accessibility
+      charitable-grants-for-ill-or-disabled-people
+      shopping-around-for-disability-aids-and-equipment
+      motability-blue-badge-scheme-and-discounted-travel
+      money-saving-tips-and-discounts-for-disabled-people
+      support-to-help-you-study-when-ill-or-disabled
+      getting-a-mortgage-if-youre-ill-or-disabled
+      getting-a-loan-if-you-are-ill-or-disabled
+      early-retirement-because-of-illness-or-disability
+      buying-insurance-if-youre-ill-or-disabled
+      long-term-care-services-at-a-glance
+      planning-ahead-for-when-you-cant-manage-your-money
+      care-services-to-help-you-stay-in-your-own-home
+      preparing-for-illness-old-age-and-death
+      care-home-or-home-care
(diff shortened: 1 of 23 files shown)


spec/helpers/sticky_footer_visibility_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    %w(en cy).each do |locale|
+      it 'excludes in any language' do
+        allow(request).to receive(:url) { "http://example.com/#{locale}/categories/insurance" }
+        allow(request).to receive(:base_url) { 'http://example.com' }
+
+        expect(display_sticky_newsletter_on_this_slug?).to be false
+      end
+    end
+  end
+end
```
tomas-stefano: All this tests can be moved the allow expectations to a before clause and being divide using lets. Or instead of using before you can add the expectation to the double and using lets too.

Tr4pSt3R: okay i will refactor that now

app/helpers/sticky_newsletter_visibility.rb
```diff
@@ -0,0 +1,15 @@
+module StickyNewsletterVisibility
+  def display_sticky_newsletter_on_this_slug?
+    @slug = request.url.gsub request.base_url, ''
```
tomas-stefano: A personal preference, I would move all the logic of this module to a class, because if you don't the @slug instance variable will be shared with all the controllers, where is possible to cause awkward bugs.

Example:

```
2.1.2 :001 > module Foo
2.1.2 :002?>   def foo
2.1.2 :003?>     @foo = 'foo'
2.1.2 :004?>     end
2.1.2 :005?>   end
 => :foo
2.1.2 :006 > class Bar
2.1.2 :007?>   include Foo
2.1.2 :008?>   end
 => Bar
2.1.2 :009 > bar = Bar.new
 => #<Bar:0x007f99f338fbf0>
2.1.2 :010 > bar.foo
 => "foo"
2.1.2 :011 > bar.instance_variable_get(:@foo)
 => "foo"
```

The @foo variable is using on the current Bar instance. If other method is assigning @slug too, now you will have a bug hard to find.

nmiddleweek: I would also prefer that we create a Component Class for this so we can keep all of the Sticky Newsletter UI Component's logic in one place.

There is GRASP Design Pattern called "Information Expert" which basically is about deciding where we put code/logic that responsible for certain software 'things'.

If we think about this, there is a strong case that the responsibility of deciding to show or hide the StickyNewsletter UI Component belongs inside a Class perhaps called StickyNewsletterUIComponent.

This would remove this code from the ApplicationController and keeps things nice and tidy.

The same should apply to the Inline Newsletter code and it's 'slug', currently this is also mixed up with the ApplicationController.

nmiddleweek: I don't know if this library would be useful to use to help with this or perhaps it's overkill for our needs: https://github.com/apotonick/cells

This was briefly discussed/ mentioned last week on Slack.

darokel: Think that needs a higher discussion outside of this PR :)

Tr4pSt3R: @nmiddleweek this file contains the vital details we need about the Sticky Newsletter. It houses all that we need. 

(https://github.com/moneyadviceservice/frontend/blob/6737_sticky-footer-visibility/features/support/ui/sections/sticky_newsletter.rb

```
require_relative '../section'

module UI::Sections
  class StickyNewsletter < UI::Section
    element :subscription_email, "input[name='subscription[email]']"
    element :send_me_money_advice, ".button--newsletter"
    element :close_button, '.news-signup-sticky__close'
  end
end
```

darokel: I think Nick was referring to encapsulating the business logic of UI components in one place. The file you're referring to is just a UI object for testing purposes.

Tr4pSt3R: @darokel oh i see what you mean. Gracias

spec/helpers/sticky_footer_visibility_spec.rb
Tr4pSt3R: @tomas-stefano  i need your help for this please

tomas-stefano: One example (I didnt test, just a draft):

```ruby
RSpec.describe StickyNewsletterVisibility, type: :helper do
  let(:request) { double('request', url: url, base_url: base_url)}

  describe 'home page' do
    %w(en cy).each do |locale|
      let(:url) { "http://example.com/#{locale}" }
      
      it 'excludes in any language' do
        expect(display_sticky_newsletter_on_this_slug?).to be false
      end
    end
  end

  describe 'category slugs' do
    %w(en cy).each do |locale|
      let(:url) { "http://example.com/#{locale}/categories" }
      
      it 'excludes in any language' do
        expect(display_sticky_newsletter_on_this_slug?).to be false
      end
    end
  end
end
```

app/helpers/sticky_newsletter_visibility.rb
tomas-stefano: We already using decorators/presenter pattern. We can use here too.

spec/helpers/sticky_footer_visibility_spec.rb
Tr4pSt3R: Hey @tomas-stefano yay! It works. Checkout the changes here: https://github.com/moneyadviceservice/frontend/commit/7b6d6d5dc5aaa05536624aa3b67f552cf9bb4e80

tomas-stefano: Cool. Now it only left the module improvements :)

spec/helpers/sticky_footer_visibility_spec.rb
```diff
@@ -0,0 +1,92 @@
+RSpec.describe StickyNewsletterVisibility, type: :helper do
+  let(:request) { double('request', url: url, base_url: base_url) }
+  let(:base_url) { "http://example.com"  }
+
```
tomas-stefano: Should have a describe ‘#name-of-the-method-being-tested' here

app/helpers/sticky_newsletter_visibility.rb
```diff
@@ -0,0 +1,31 @@
+module StickyNewsletterVisibility
+  class StickyNewsletterFilter
+
+    BLACKLIST_FILTERED = [ '', 'categories', 'corporate', 'tools', 'videos','corporate_categories', 'sitemap', 'users']
```
tomas-stefano: Can you split into lines here, please?

spec/helpers/sticky_footer_visibility_spec.rb
Tr4pSt3R: @tomas-stefano do you mean here? 
```
RSpec.describe StickyNewsletterVisibility, type: :helper do
```

tomas-stefano: No, I mean describe '#method-that-you-are-testing'. Wrap it :)

spec/helpers/sticky_footer_visibility_spec.rb
```diff
@@ -0,0 +1,92 @@
+RSpec.describe StickyNewsletterVisibility, type: :helper do
```
tomas-stefano: Other thing that I realize. Can you rename this file to `sticky_newsletter_visibility_spec.rb`?

Tr4pSt3R: yeah shall do so

Tr4pSt3R:
Hi @darokel could you please have another look? The code has been dry'd up with shared_examples

darokel:
:+1:

rnalexander:
:+1: :shipit:

PR #1314, opened 2015-12-04
Feedback Panel Feature Flag

darokel:
This UI panel has recently resurfaced due to removing the feature flags. We’re putting it behind a feature flag to keep it hidden until we get a decision on whether it’s needed anymore or not.

### DIFF
--- a/app/views/shared/_feedback.html.erb
+++ b/app/views/shared/_feedback.html.erb
@@ -1,9 +1,15 @@
-<section class="panel feedback-panel">
-  <h2 class="feedback-panel__title"><%= t('feedback_panel.title') %></h2>
-  <p><%= t('feedback_panel.greeting') %></p>
-  <ul>
-    <li><%= link_to t('feedback_panel.article_feedback_link_text'), new_article_feedback_path(article.object.id), class: 't-feedback-panel__article-feedback-link' %></li>
-    <li><%= link_to t('feedback_panel.technical_feedback_link_text'), new_technical_feedback_path, class: 't-feedback-panel__technical-feedback-link' %></li>
-    <li><%= link_to t('feedback_panel.advice_link_text'), advice_path, class: 't-feedback-panel__advice-link' %></li>
-  </ul>
-</section>
+<% if Feature.active?(:feedback) %>
+  <section class="panel feedback-panel">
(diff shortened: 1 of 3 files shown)


features/step_definitions/feedback_steps.rb
```diff
@@ -16,7 +16,8 @@
 end
 
 Then(/^I should see the feedback panel$/) do
-  expect(article_page).to have_feedback_panel
+  pending
```
Tr4pSt3R: it doesn't need to be pending as it won't be run as per ```@enable-feedback```.

darokel: Not sure what you mean but it does run. This has been marked pending because there's a weird glitch somewhere causing the feature flag not to be enabled for this scenario. It works fine in the browser. Looking into the issue is outside of this PR as we don't know whether it'll stay or not and this change needs to get out of the door before the weekend.

Tr4pSt3R:
:+1: :sheep: it

tomas-stefano:
:+1:

features/step_definitions/feedback_steps.rb
asmega: partially figured it out. this snippet will offer insight. this was whilst in the view hence the view never rendered the feedback panel

```
[1] pry(#<#<Class:0x007fd97ad349a0>>)> Feature.instance_variable_get :@active_features
=> [:rio,
 :budget_planner,
 :agreements,
 :improvements,
 :car_cost_tool,
 :pensions_calculator,
 :timelines,
 :cutback_calculator,
 :pensions_and_retirement,
 :debt_test,
 :action_plans,
 :debt_advice_locator,
 :health_check,
 :previous_and_next_links,
 :popular_links,
 :corporate,
 :profile,
 :cost_calculator_builder,
 :annuities_landing_page,
 :quiz,
 :savings_calculator,
 :cms_home_page,
 :feedback]
[2] pry(#<#<Class:0x007fd97ad349a0>>)> Feature.active? :feedback
=> false
[3] pry(#<#<Class:0x007fd97ad349a0>>)>
```

darokel:
I've fixed the bug with the Feature gem. Removed the pending status

PR #1317, opened 2015-12-07
[DEV APPROVED] Updating Hero CSS

johnplayer1982:
Updating Hero slot CSS, removing references to images as they will be served inline from CMS.


johnplayer1982:
full width
posh tablet
budget tablet
fancy hi res phone
my cheap phone

Tr4pSt3R:
:+1:

slaywell:
:+1: 

@johnplayer1982 as discussed let's keep the double-density version of the image in the CMS and add an alternative layout for IE8.

asmega:
:+1:

PR #1318, opened 2015-12-08
update xml sitemaps

asmega:


Tr4pSt3R:
:+1:

public/category-sitemap.xml
```diff
@@ -1,299 +1,310 @@
-<?xml version="1.0" encoding="utf-8"?>
```
munckymagik: Looks like the `<?xml ...?>` tag has disappeared from all of these docs. Is that deliberate? Otherwise :+1:

tomas-stefano:
:+1:

public/category-sitemap.xml
asmega: these were the files that were handed over to me. they passed on the w3c validator with a few warnings, so appear to be valid

Tr4pSt3R:
:shipit: :ship:

public/category-sitemap.xml
munckymagik: Okey dokey.

munckymagik:
:+1:

PR #1319, opened 2015-12-09
Social share functionality generic

davidtrussler:
This PR takes a bunch of styles removed from Dough (PR #244) which are now more generic and implements them in frontend, adding MAS-specific styles.


slaywell:
As discussed, please can you move the `.icon` styles back into the _icon.scss file to follow our BEM conventions?

davidtrussler:
The problem with that change is that _social_sharing.scss needs the variables defined in _icon.scss which could cause problems. That is why the two separate files were created.

slaywell:
Either:

app/assets/stylesheets/components/common/_social_sharing.scss
```diff
@@ -0,0 +1,124 @@
...
+  padding: ($social-sharing-icon-height - $twitter-icon-height) / 2 ($social-sharing-icon-width - $twitter-icon-width) / 2;
+  background: $color-twitter;
+
+  &:hover {
+    background: $color-twitter-hover;
+  }
+}
+
+.social-sharing__item--email {
+  padding: ($social-sharing-icon-height - $email-icon-height) / 2 ($social-sharing-icon-width - $email-icon-width) / 2;
```
slaywell: As discussed this line is probably a bit over-elaborate for what it needs to do, and changing this should remove the need for the shared variable

slaywell:
Looks good! :+1:

tomas-stefano:
:+1:

PR #1320, opened 2015-12-09
[WIP] Fixing tool newsletter button style

johnplayer1982:
A fix for the newsletter button appearing grey in tools which use Dough:

before

This happens because frontend loads dough then applies its own css properties to the newsletter button.  Then the tool will load dough again and apply the default button style, overwriting the custom properties outlined in frontend.

My proposed fix is to remove the classname 'button' from the newsletter button tag, and then extend .button within the class (.button--newsletter):

Outcome:
after

This will fix all tools with this issue

Fix also applied to sticky footer:

screen shot 2015-12-11 at 10 06 10

After:

### DIFF
--- a/app/assets/stylesheets/components/dough_theme/button/_button.scss
+++ b/app/assets/stylesheets/components/dough_theme/button/_button.scss
@@ -122,34 +122,3 @@
   @extend %type-button-small;
   padding: 14px 16px;
 }
-
-// Lives in newsletter signup module
-//
-// .button--newsletter
-//
-// Styleguide Newsletter signup buttons
-
-.button--newsletter {
-  @include body(12,14);
-  background: $color-newsletter;
-  border-bottom: 0;
-  color: $color-white;
-  font-weight: 500;
-  text-transform: uppercase;
-
-  &:focus,
-  &:hover,
-  &:active {
-    color: $color-true-black;
-    background: $color-newsletter-active;
-    outline: none;
-    border-top: 0;
-    padding-top: $button-padding-vertical;
-  }
-
-  &:visited,
-  &:active {
-    background: $color-newsletter;
-    color: $color-white;
-  }
-}
--- a/app/views/shared/_newsletter_signup.html.erb
+++ b/app/views/shared/_newsletter_signup.html.erb
@@ -25,7 +25,7 @@
       ) %>
       <%= hidden_field_tag 'subscription[ga_categoryid]', 'Newsletter SignUp' -%>
       <%= hidden_field_tag 'subscription[category]', 'footer' -%>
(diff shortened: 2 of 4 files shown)


app/views/shared/_newsletter_signup.html.erb
```diff
@@ -25,7 +25,7 @@
       ) %>
       <%= hidden_field_tag 'subscription[ga_categoryid]', 'Newsletter SignUp' -%>
       <%= hidden_field_tag 'subscription[category]', 'footer' -%>
-      <%= button_tag(t('newsletter_subscriptions.button'), type: 'submit', class: 'newsletter-button button button--newsletter t-newsletter-button') %>
+      <%= button_tag(t('newsletter_subscriptions.button'), type: 'submit', class: 'newsletter-button button--newsletter t-newsletter-button') %>
```
nmiddleweek: I don't think this ```newsletter-button``` class is doing anything?

app/assets/stylesheets/components/dough_theme/button/_button.scss
```diff
@@ -130,6 +130,7 @@
 // Styleguide Newsletter signup buttons
 
 .button--newsletter {
+  @extend .button;
```
nmiddleweek: You probably meant to extend . newsletter-button from .button or rename .button-newsletter to .newsletter-button ?

app/views/shared/_newsletter_signup.html.erb
johnplayer1982: It only adds the min-width, which could probably be moved into button--newsletter

app/assets/stylesheets/components/dough_theme/button/_button.scss
johnplayer1982: I've renamed .button-newsletter to .newsletter-button, that makes more sense

davidtrussler:
:+1:

slaywell:
Looks good except having .newsletter-button in our _buttons.scss file isn't following our BEM / file naming conventions. Can we move it to the _newsletter_signup.scss file and change the class name to .newsletter-signup__something?

app/assets/stylesheets/_enhanced.scss
```diff
@@ -31,10 +31,10 @@
 @import 'layout/page_specific/all';
 
 // Components
-@import 'components/common/all';
 @import 'components/page_specific/all';
 @import 'components/dough_theme/variables';
 @import 'components/dough_theme/*/_all.scss';
+@import 'components/common/all';
```
johnplayer1982: For this to work I had to move the component CSS import so they are loaded after the dough_theme variables, as there is a variable which is used by the newsletter button which is defined in dough

PR #1321, opened 2015-12-09
[DEV APPROVED] 6797 adding in contact panel borders

slaywell:
In a previous PR for ticket 6797 we accidentally removed some borders on the contact panels. This PR adds them back in but also adds a flexible height to the borders so that they will increase in height when the content is longer.


johnplayer1982:
:+1:

davidtrussler:
:+1:

PR #1339, opened 2015-12-21
Remove Sticky Footer from survive january campaign.

nmiddleweek:


Tr4pSt3R:
:+1:

Tr4pSt3R:
Although, it should be tested. This "Survive January" will be removed after January right?

johnplayer1982:
:+1:

slaywell:
@Tr4pSt3R Yes - it will be removed. We will get a ticket added to the backlog.

PR #1345, opened 2016-01-04
[DEV APPROVED] #6793 Accessibility updates to the sticky footer

slaywell:
Addressing the following accessibility recommendations (see TP ticket for details) 

* Adding aria roles for dialog and labelledby
* Setting keyboard focus when it loads
* Creating a closed tab loop for the fields within it
* Returning keyboard focus to the page when it is closed

### DIFF
--- a/app/assets/javascripts/components/NewsletterSticky.js
+++ b/app/assets/javascripts/components/NewsletterSticky.js
@@ -53,6 +53,10 @@ define([
     this.closeButton = this.$el.find(this.config.closeClassSelector);
     $window = $(window);
 
+    var focusable = this.$el.find('a[href], area[href], input:not([disabled], [type=hidden]), select:not([disabled], [hidden]), textarea:not([disabled], [hidden]), button:not([disabled], [hidden]), iframe, object, embed, *[tabindex], *[contenteditable]');
+    this.firstFocusElement = focusable.first();
+    this.lastFocusElement = focusable.last();
+
     $newsletterStickForm = this.$el.find('form');
     newsletterCloseBoxCookieURL = $newsletterStickForm.find('#close-box-cookie-url').val();
   };
@@ -66,10 +70,11 @@ define([
 
     if (isActive) {
       this.closeButton.on('click', $.proxy(this._handleCloseClick, this));
+      this.firstFocusElement.on('keydown', $.proxy(this._handleFirstElementTab, this));
+      this.lastFocusElement.on('keydown', $.proxy(this._handleLastElementTab, this));
       this.$el.on('click', this.config.buttonDoneClass, $.proxy(this._handleCloseClick, this));
       $newsletterStickForm.on('ajax:error', $.proxy(this._formAjaxErrorHandler, this));
-    }
-    else {
+    } else {
       this.closeButton.off('click', $.proxy(this._handleCloseClick, this));
       this.$el.off('click', this.config.buttonDoneClass, $.proxy(this._handleCloseClick, this));
       $newsletterStickForm.off('ajax:error', $.proxy(this._formAjaxErrorHandler, this));
@@ -112,6 +117,9 @@ define([
   NewsletterSticky.prototype._handleScroll = function() {
     if (this._hasScrolledOverPercentage(this.config.appearsAfterPercentage)) {
       this.$el.addClass(this.config.visibleClassName);
+
+      this.previousFocusElement = $(':focus');
+      this.firstFocusElement.focus();
       this._setScrollListener(false);
 
       MAS.publish('analytics:trigger', {
@@ -218,7 +226,25 @@ define([
     this.$el.removeClass(this.config.visibleClassName);
     this.closeButton.addClass(this.config.closedClassName);
     this._setListeners(false);
+    this.previousFocusElement.focus();
   };
+
+  /*
+  * Handle the tab event on the last and first focusable element
+  * A tab loop is created until the user dismisses the dialog (accessibility recommendation)
+   */
+  NewsletterSticky.prototype._handleLastElementTab = function(e) {
+    if (e.which === 9 && !e.shiftKey) {
+      e.preventDefault();
+      this.firstFocusElement.focus();
+    }
+  }
+  NewsletterSticky.prototype._handleFirstElementTab = function(e) {
+    if (e.which === 9 && e.shiftKey) {
+      e.preventDefault();
+      this.lastFocusElement.focus();
+    }
+  }
 
   return NewsletterSticky;
 });
--- a/app/assets/javascripts/components/NewsletterSticky.js
+++ b/app/assets/javascripts/components/NewsletterSticky.js
@@ -18,7 +18,8 @@ define([
     },
     $newsletterStickForm,
     newsletterCloseBoxCookieURL,
-    $window;
+    $window,
+    TAB_KEY = 9;
 
   NewsletterSticky = function($el, config) {
     NewsletterSticky.baseConstructor.call(this, $el, config, defaultConfig);
@@ -50,14 +51,15 @@ define([
    * Sets up local variables for the sticky newsletter component
    */
   NewsletterSticky.prototype._setVars = function() {
-    this.closeButton = this.$el.find(this.config.closeClassSelector);
-    $window = $(window);
-
     var focusable = this.$el.find('a[href], area[href], input:not([disabled], [type=hidden]), ' +
       'select:not([disabled], [hidden]), textarea:not([disabled], [hidden]), ' +
       'button:not([disabled], [hidden]), iframe, object, embed, *[tabindex], *[contenteditable]');
-    this.firstFocusElement = focusable.first();
-    this.lastFocusElement = focusable.last();
+
+    this.closeButton = this.$el.find(this.config.closeClassSelector);
+    $window = $(window);
+    
+    this._firstFocusElement = focusable.first();
+    this._lastFocusElement = focusable.last();
 
     $newsletterStickForm = this.$el.find('form');
     newsletterCloseBoxCookieURL = $newsletterStickForm.find('#close-box-cookie-url').val();
@@ -72,8 +74,8 @@ define([
 
     if (isActive) {
       this.closeButton.on('click', $.proxy(this._handleCloseClick, this));
-      this.firstFocusElement.on('keydown', $.proxy(this._handleFirstElementTab, this));
-      this.lastFocusElement.on('keydown', $.proxy(this._handleLastElementTab, this));
+      this._firstFocusElement.on('keydown', $.proxy(this._handleFirstElementTab, this));
+      this._lastFocusElement.on('keydown', $.proxy(this._handleLastElementTab, this));
       this.$el.on('click', this.config.buttonDoneClass, $.proxy(this._handleCloseClick, this));
       $newsletterStickForm.on('ajax:error', $.proxy(this._formAjaxErrorHandler, this));
     } else {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -53,6 +53,10 @@ define([
     this.closeButton = this.$el.find(this.config.closeClassSelector);
     $window = $(window);
 
+    var focusable = this.$el.find('a[href], area[href], input:not([disabled], [type=hidden]), select:not([disabled], [hidden]), textarea:not([disabled], [hidden]), button:not([disabled], [hidden]), iframe, object, embed, *[tabindex], *[contenteditable]');
```
munckymagik: This seems a bit long. What's the MAS line width standard for JS?

slaywell: We follow the airbnb style guide for Js, which has jshint config for limiting it to 100 characters. So yep, bit long - I'll change this.

munckymagik: For the Ruby at least I think it's 120 characters. Couldn't find anything about it in https://github.com/moneyadviceservice/frontend-code-standards directly, however it refers to the AirBnB standard, which says 100 chars (https://github.com/airbnb/javascript#6.2)

munckymagik: Ok you got there before me :)

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -218,7 +226,25 @@ define([
...
     this._setListeners(false);
+    this.previousFocusElement.focus();
   };
 
+  /*
+  * Handle the tab event on the last and first focusable element
+  * A tab loop is created until the user dismisses the dialog (accessibility recommendation)
+   */
+  NewsletterSticky.prototype._handleLastElementTab = function(e) {
+    if (e.which === 9 && !e.shiftKey) {
```
munckymagik: Does `e.which === 9` identify the tab key?

slaywell: Yep

slaywell:
Adding the Accessibility Centre's suggestions for context:

slaywell:
Adding the Accessibility Centre's comments for context:

*I think the role=dialog is your best bet. I have found some information from Marco Zehe’s blog, which explains this a lot better, and what should be done. However I don’t think that you are going to please everyone anyway, as there is obviously a change in context, which all users hate.*
 
* *If you use the role “dialog”, you must use aria-labelledby to point to the element containing the visible dialog title. The aria-labelledby attribute must be on the same HTML element as the role “dialog” attribute.*
* *If your dialog has one main descriptive text, you must use aria-describedby to point to its element, also on the same element that has role “dialog”. You can also point to multiple text elements by separating their IDs with spaces.*
 
*Also make sure your keyboard focus never leaves the dialog. A few rules for this:*
 
* *Create a natural tab order by arranging elements in the DOM in a logical order. That will cause the tab order to flow naturally without you having to do much. If you have to use the tabIndex attribute for anything, give it a value of 0 so the element fits into the natural tab order.*
* *On the very last element of your dialog, add a special keyPress handler that traps the forward tab key to redirect it to the first element of the dialog. Otherwise, tab would jump outside the dialog into the surrounding document or to the browser UI, and we don’t want that.*
* *Likewise, add a keyPress handler to the very first focusable element inside the dialog that traps the shift+tab key combination so the backwards tab order also doesn’t leave the dialog.*
 
*Oh and always make sure that you set focus back to where you were for keyboard users. This is _VERY_ important.*

nmiddleweek:
I don' think this branch has been updated with the lastest-master branch. It's always good practice to do a rebase from master before the PR.

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -53,6 +53,12 @@ define([
     this.closeButton = this.$el.find(this.config.closeClassSelector);
     $window = $(window);
 
+    var focusable = this.$el.find('a[href], area[href], input:not([disabled], [type=hidden]), ' +
```
nmiddleweek: It would be good to move this var declaration to the top of the function.

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -53,6 +53,12 @@ define([
...
+      'select:not([disabled], [hidden]), textarea:not([disabled], [hidden]), ' +
+      'button:not([disabled], [hidden]), iframe, object, embed, *[tabindex], *[contenteditable]');
+    this.firstFocusElement = focusable.first();
+    this.lastFocusElement = focusable.last();
```
nmiddleweek: Strictly speaking these are private properties so an underscore prefix helps indicate that. ```this._firstFocusElement```

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -112,6 +119,9 @@ define([
   NewsletterSticky.prototype._handleScroll = function() {
     if (this._hasScrolledOverPercentage(this.config.appearsAfterPercentage)) {
       this.$el.addClass(this.config.visibleClassName);
+
+      this.previousFocusElement = $(':focus');
```
nmiddleweek: This effectively does this ```$('*:focus')``` which will scan the DOM, I wonder if using ```document.activeElement``` is as widely accepted?

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -218,7 +228,26 @@ define([
...
+    this.previousFocusElement.focus();
   };
 
+  /*
+  * Handle the tab event on the last and first focusable element -
+  * a tab loop is created until the user dismisses the dialog (accessibility recommendation).
+  * The keycode for the tab key is 9.
+   */
+  NewsletterSticky.prototype._handleLastElementTab = function(e) {
+    if (e.which === 9 && !e.shiftKey) {
```
nmiddleweek: It might be worth setting a TAB_KEY var after $window on line 20 just so it's super clear in the future without having to look up the ASCII codes...

```
.
.
.
$newsletterStickForm,
newsletterCloseBoxCookieURL,
$window,
TAB_KEY = 9;
```

munckymagik:
:+1:

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -66,10 +74,11 @@ define([
...
     if (isActive) {
       this.closeButton.on('click', $.proxy(this._handleCloseClick, this));
+      this._firstFocusElement.on('keydown', $.proxy(this._handleFirstElementTab, this));
+      this._lastFocusElement.on('keydown', $.proxy(this._handleLastElementTab, this));
       this.$el.on('click', this.config.buttonDoneClass, $.proxy(this._handleCloseClick, this));
       $newsletterStickForm.on('ajax:error', $.proxy(this._formAjaxErrorHandler, this));
-    }
-    else {
+    } else {
       this.closeButton.off('click', $.proxy(this._handleCloseClick, this));
```
nmiddleweek: I missed this first time round... We need to turn the handlers off to free up memory...

```this._firstFocusElement.off('keydown', $.proxy(this._handleFirstElementTab, this));
this._lastFocusElement.off('keydown', $.proxy(this._handleLastElementTab, this));```

File: spec/javascripts/fixtures/NewsletterSticky.html [hunk 8]
```diff
@@ -1,9 +1,11 @@
 <div id="fixture-1">
   <div class="news-signup-sticky news-signup-sticky--positioned" data-dough-component="NewsletterSticky">
     <form onsubmit="return false;">
-      <button class="news-signup-sticky__close">
+      <button class="news-signup-sticky__close" id="first-focusable">
```
nmiddleweek: That's a little risky for an ID, if this test pulled in multiple fixtures both having that ID... Albeit it's a very slim chance so it's debatable.

spec/javascripts/fixtures/NewsletterSticky.html
```diff
@@ -1,9 +1,11 @@
...
           &times;
       </button>
+      <input type="text" id="second-focusable">
+      <input type="text" id="third-focusable">
```
nmiddleweek: dito

spec/javascripts/fixtures/NewsletterSticky.html
mconnell: There's a pattern within the ruby applications (certainly within RAD and codebases which use site prism) which have javascript tests to us classes with a 't-' prefix, eg.

https://github.com/moneyadviceservice/rad/blob/master/spec/javascripts/tests/MultiTableFilter_spec.js#L9
https://github.com/moneyadviceservice/rad/blob/master/spec/javascripts/fixtures/MultiTableFilter.html#L19

I think the intention is so that in the event of ids / classes being refactored/removed, it's more obvious that the particular element is under test by something. I don't know if this is a pattern within the frontend codebase, but it's maybe something that could be a discussion point at the next frontend meeting?

app/assets/javascripts/components/NewsletterSticky.js
```diff
@@ -218,7 +231,27 @@ define([
...
+  * a tab loop is created until the user dismisses the dialog (accessibility recommendation).
+   */
+  NewsletterSticky.prototype._handleLastElementTab = function(e) {
+    if (e.which === TAB_KEY && !e.shiftKey) {
+      e.preventDefault();
+      this._firstFocusElement.focus();
+    }
+  }
+  NewsletterSticky.prototype._handleFirstElementTab = function(e) {
+    if (e.which === TAB_KEY && e.shiftKey) {
```
mconnell: #TodayILearned There's a shiftKey function on Event.

mconnell:
:+1:

spec/javascripts/fixtures/NewsletterSticky.html
nmiddleweek: I think ```t-``` prefix is for CSS themes and ```qa-``` is used for qa/test hooks in a similar way that ```js-``` prefix is for javascript hooks (more generally on CSS classes and not IDs). However even with any of those prefixes applied, the issue still stands. To reduce the risks though for this fixture, it could be good to rename the ID to ```X-news-signup-sticky__first-focusable``` or safer still, delete the ID and add the class ```X-news-signup-sticky__first-focusable``` (where X is the prefix we decide).

davidtrussler: I understand that t- represents test and shouldn't really be used for styling. Often these classes shouldn't really still be there but have tended to remain after the testing has occurred.

munckymagik: The `t-*` MUST NOT be removed or the automated test suites will break. Sorry to seem shouty here, but they are not part of a one-off test cycle.

They are vital hooks into the HTML/view layer for automated end to end tests that stop us having to depend on the structure of the DOM to find the things we need to check. This makes the test less brittle as changes to the visual design are less likely to break the tests.

Please don't remove these unless you can be sure there are no references to them anywhere under the `/spec` folder of a project.

munckymagik: Only `l-` and `js-` are defined in our styleguide (https://www.moneyadviceservice.org.uk/en/styleguide/css). The `t-` has been heavily used across all the MAS project for test hooks and I've not seen anything that uses them for theme styles.

darokel: Nick we discussed last week that t prefixes are only used for testing not styling?

davidtrussler: @munckymagik Point taken on t- prefixes - no problem with shouty-ness!

slaywell:
Didn't actually need the IDs in the fixture. @nmiddleweek could you check the removal of the handlers please?

app/assets/javascripts/components/NewsletterSticky.js
slaywell: Done!

nmiddleweek:
nice one :+1:

PR #1357, opened 2016-01-14
Set the domain for webchat using a value in Puppet

misaka:
One of the outcomes of the recent problems with WebChat was that this value is hard-coded in the front-end when we actually have pre-environment values set in Puppet. This change allows us to use the values in Puppet.

Tested in preview.


slaywell:
:+1:

jongilbraith:
:+1:

PR #1362, opened 2016-01-20
Bump debt advice locator version to 2.0.4.97

johnplayer1982:
Bumping debt advice locator tool

Relates to PR's:

1) #6918 Contact number adjustments:
https://github.com/moneyadviceservice/debt-advice-locator/pull/105

2)#6918 Changing phone number colour
https://github.com/moneyadviceservice/debt-advice-locator/pull/107

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -240,7 +240,7 @@ GEM
       sass-rails (~> 4.0.0)
       uglifier
     database_cleaner (1.3.0)
-    debt_advice_locator (2.0.4.94)
+    debt_advice_locator (2.0.4.97)
       active_attr (~> 0.8)
       active_model_serializers (= 0.8.2)
       addressable


Gemfile.lock
```diff
@@ -557,7 +557,7 @@ GEM
       thor (>= 0.18.1, < 2.0)
     rainbow (2.0.0)
     raindrops (0.13.0)
-    rake (10.4.2)
+    rake (10.5.0)
```
johnplayer1982: Would be good to know if I need this update? It came with running 'bundle update debt_advice_locator' to update the version

munckymagik: The official changes between 10.4.2 and 10.5 look pretty safe to me:

* [Rake History.rdoc](https://github.com/ruby/rake/blob/2ffe31d0dfb59ef00f3af5f1c27fe33153c01532/History.rdoc)
* [10.4.2 to 10.5.0 comparison](https://github.com/ruby/rake/compare/v10.4.2...v10.5.0)

If you can still run rake without any errors then I'd say it's fine e.g. try `bundle exec rake routes`.

What I do recommend is, to split the two concerns here. Split the change to put the rake update in a separate commit, that way if someone wants/needs to revert that ever it can be done with `git revert` and it makes it very visible in the log when the change happened.

My 2p.

johnplayer1982: Thanks @munckymagik I agree with that, tests pass fine but like you say it will be easier to roll back if necessary if its an isolated commit

Tr4pSt3R:
:+1:

munckymagik:
:+1:

PR #1365, opened 2016-01-22
[DEV APPROVED] id: 6920, comment: Adds sticky header to mobile site with some exceptions

davidtrussler:
-  Adds CSS to fix position of header on smaller screen sizes
- restyles header globally to make it less deep
- updates main content position in line with changes to header depth
- Add non-sticky class when on a tool page (with the exception of payday loans tool, which has own layout)
- Adds styles to remove sticky header from selected pages

Integrates with PR239 on mortgage_calculator

### DIFF
--- a/app/assets/stylesheets/components/common/_mas_logo.scss
+++ b/app/assets/stylesheets/components/common/_mas_logo.scss
@@ -2,6 +2,7 @@
   margin-left: -100%;
   max-width: none;
   width: 200%;
+  display: block;
 }
 
 .mas-logo__link {
@@ -9,8 +10,6 @@
   overflow: hidden;
   vertical-align: middle;
   width: 200px;
-  font-size: 0;
-  line-height: 1;
 
   @include respond-to($mq-m) {
     width: 285px;
--- a/app/assets/stylesheets/components/common/_mas_logo.scss
+++ b/app/assets/stylesheets/components/common/_mas_logo.scss
@@ -9,6 +9,8 @@
   overflow: hidden;
   vertical-align: middle;
   width: 200px;
+  font-size: 0;
+  line-height: 1;
 
   @include respond-to($mq-m) {
     width: 285px;
--- a/app/assets/stylesheets/layout/common/_header.scss
+++ b/app/assets/stylesheets/layout/common/_header.scss
@@ -1,7 +1,7 @@
 .l-header {
-  background: $color-green-primary;
-  min-height: $baseline-unit*11;
-  position: relative;
+  width: 100%;
+  z-index: 200;
+  min-height: 55px;
 
   .mobile-nav {
     display: none;
@@ -9,13 +9,9 @@
 
   .mas-logo {
     @include column(4);
-    margin-bottom: $baseline-unit*3;
-    margin-top: $baseline-unit*3;
+    margin-bottom: $baseline-unit*2;
+    margin-top: $baseline-unit*2;
     position: relative;
-
-    @include respond-to($mq-small-tablet) {
-      margin-bottom: $baseline-unit*3;
-    }
 
     @include respond-to($mq-m) {
       @include column(10);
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/components/common/_mas_logo.scss
```diff
@@ -9,6 +9,8 @@
   overflow: hidden;
   vertical-align: middle;
   width: 200px;
+  font-size: 0;
+  line-height: 1;
```
slaywell: Display:block on the image might be neater?

app/assets/stylesheets/layout/common/_header.scss
```diff
@@ -1,7 +1,7 @@
 .l-header {
-  background: $color-green-primary;
```
slaywell: Just moving the color to the inner element is problematic when you get to Desktop width, you no longer have a full-width header.

johnplayer1982: Might be worth adding this to the header element CSS instead if the layout container within it

app/assets/stylesheets/layout/common/_header.scss
```diff
@@ -1,7 +1,7 @@
...
-  min-height: $baseline-unit*11;
-  position: relative;
+  width: 100%;
+  z-index: 200;
+  min-height: 55px;
```
slaywell: Can we use baseline-unit?

app/assets/stylesheets/layout/common/_header.scss
```diff
@@ -180,6 +180,24 @@
   }
 }
 
+.l-header__content {
+  position: fixed;
+  background: $color-green-primary;
+  z-index: 150;
```
slaywell: Is there any way we can make keep this and the other z-index on line 3 under 100? Based on what we saw yesterday with the Dragon browsing tech?

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: I agree that would be better but the problem is that we are mixing units (px and rem) which we could do with calc() in CSS which is not generally used on the site. There's also the px() SASS function which converts px to em which may be exploitable somehow.

slaywell: Not sure I follow, what was wrong with `min-height: $baseline-unit*11;` that was there before (or just using a higher multiple than 11)?

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: Problem here is that the inner element is what is sticky whilst the header itself scrolls off the viewport in this scenario, which we implemented to retain the depth of the header when its position is fixed.

slaywell: I think that putting the green background color on both elements will fix it?

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: The actual height of the header is 30px + 1.5rem, i.e. is still a mix of pixels and rems so whether that is expressed in pixels or rems seems to me not to make much difference - still a bit imperfect either way.

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: The value for this needs to be greater than that for the sticky newsletter signup, which is 100, so that would need to change as well. Might be better left out of this card and added as a more generic piece of work.

slaywell: Sounds good, I thought this might be the case!

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: The media query fix is better I think because of the px/rem issue for the header height means there are potentially *slight* variations in the depth of the two elements.

app/assets/stylesheets/layout/common/_header.scss
davidtrussler: Nonetheless I'll define the header height as $baseline-unit*9 (which is the equivalent according to my calculations) for the smaller viewport in order to maintain continuity.

johnplayer1982:
Pulled locally and tested, it gets a :+1: from me

slaywell:
:+1:

PR #1366, opened 2016-01-22
[DEV APPROVED] Removed baby cost calc

habitullence:


Tr4pSt3R:
:+1:

.node-version
```diff
@@ -1 +1 @@
-0.10.26
+0.10.36
```
tomas-stefano: It should be included on this PR?

tomas-stefano:
:+1:

app/controllers/baby_cost_calculator_controller.rb
```diff
@@ -1,4 +1,10 @@
 class BabyCostCalculatorController < EmbeddedToolsController
+  before_action :deny_non_syndicates
+
+  def deny_non_syndicates
+    redirect_to main_app.article_path('baby-costs-calculator'), status: 301 if syndicated_tool_request? == false
```
tomas-stefano: I would use `unless syndicated_tool_request?` and then you dont need the check `== false`

jongilbraith: Or ``if !syndicated_tool_request?``, but either way would be good to do away with the explicit `==`.

spec/requests/tool_integration/baby_cost_calculator_spec.rb
```diff
@@ -1,14 +1,29 @@
...
-        get path
+    en
+    cy
+  ).each do |locale|
+    context "locale #{locale}" do
+      let (:path) { "/#{locale}/tools/" + "ToolMountPoint::BabyCostCalculator::#{locale.upcase}_ID".constantize }
+
+      describe 'non-syndicated tool request' do
+        before do
+          get path, nil, { 'X-Syndicated-Tool' => false }
```
jongilbraith: Do we still get the header at all on non syndicated requests? If not, might be appropriate to make the test mimic that.

spec/requests/tool_integration/baby_cost_calculator_spec.rb
```diff
@@ -1,14 +1,29 @@
...
+        end
+
+        it "redirects to article baby cost calculator" do
+          expect(response.status).to eq 301
+          expect(response).to redirect_to article_path('baby-costs-calculator', locale: locale)
```
Tr4pSt3R: Hey @habitullence this redirect will work but, with a hidden gotcha :)

Whereas, https://www.moneyadviceservice.org.uk/en/articles/baby-costs-calculator will work. The Welsh link https://www.moneyadviceservice.org.uk/cy/articles/baby-costs-calculator will not work. 

The correct welsh link is https://www.moneyadviceservice.org.uk/cy/articles/cyfrifiannell-costau-babi

habitullence: Good catch, I've updated the PR.

Tr4pSt3R:
Just @ me when it's pushed and I will thumb it

spec/requests/tool_integration/baby_cost_calculator_spec.rb
habitullence: You're right, header should either be there or not.

habitullence:
How do you organise your localisation files?

Tr4pSt3R:
@habitullence there are no fast rules.
 
Have a look at this simple example.
https://github.com/moneyadviceservice/frontend/blob/master/config/locales/en.yml#L371

habitullence:
Updated with I18n.

tomas-stefano:
:+1: :+1:

Tr4pSt3R:
:+1: :+1: :+1:

PR #1378, opened 2016-02-05
id: 7010, comment: Adjusts position of headings when linked from internal menu

davidtrussler:
The problem is that since the introduction of the Sticky Header on mobile devices elements that scroll to the top of the viewport when selected from internal menus are partially hidden by the header: 

image

image

This fix adds ensures there is some space above the relevant headings: 

image


nmiddleweek:
:+1:

johnplayer1982:
:+1:

PR #1385, opened 2016-02-15
[DEV APPROVED] id: 7021, comment: Switches relative positions of header and global alert on mobile views when alert is displayed

davidtrussler:
Problem is that the global alert pushes the position of the header down:
alert_before_no-scroll
====

When scrolled the header (now in fixed position) remains in this position: 
alert_before_scroll
====

The agreed solution is to switch the positions of the header and global alert in mobile view when the global alert is displayed: 
alert_after_no-scroll
====

This displays as desired when the view is scrolled: 
alert_after_scroll

### DIFF
--- a/app/assets/stylesheets/layout/common/_header.scss
+++ b/app/assets/stylesheets/layout/common/_header.scss
@@ -1,7 +1,10 @@
+// header variables
+$header-height: $baseline-unit*9;
+
 .l-header {
   width: 100%;
   z-index: 200;
-  min-height: $baseline-unit*9;
+  min-height: $header-height;
   background-color: transparent;
 
   @include respond-to($mq-m) {
@@ -188,6 +191,7 @@
 
 .l-header__content {
   position: fixed;
+  top: 0;
   background-color: $color-green-primary;
   z-index: 150;
   max-height: 100%;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/layout/common/_header.scss
```diff
@@ -184,6 +184,17 @@
       display: block;
     }
   }
+
+  // position above alert when alert is present and viewport > breakpoint
+  .global-alert + & {
+    margin-top: -4.5rem;
+    margin-bottom: 4.5rem;
+
```
slaywell: Nice use of the adjacent selector, I always forget about this.

But I don't think it's a good idea using an absolute value for the global alert size. I found an issue when you sign up, sometimes the message wraps onto two lines:

image

slaywell:
Not sure that the fix for this issue needs to be switching the position of the header and the alert. The alert appears above the header on desktop view so I think it's intentional that it's at the top of the page. 

How about disabling the header 'stickiness' when the `global-alert` message is present?

davidtrussler:
Re positioning of alert at the top of the page: this is where the alert is designed to be but in discussions with Karine it was decided to implement this solution (moving it below the header) in this case in order to retain the sticky header wherever possible, in preference to maintaining the position of the alert.

slaywell:
:+1: Looks good to me - tested a few devices and all of the alert messages I could find.

johnplayer1982:
:+1: Looks good!

PR #1392, opened 2016-02-22
Customisable footer

mconnell:
__This PR is complimented by work done in the CMS at__ https://github.com/moneyadviceservice/cms/pull/318 

* Adds the ability to read some footer content directly from the CMS.

* Wrapped in a `cms_footer` feature flag. Unless enabled, it fetches content from a `Core::CMS::Repository::Footer::Static`, in the same way the cms home page feature was implemented.


rnalexander:
How does it handle the content being in welsh?

paulanthonywilson:
:+1: 

Here for handling Welsh @rnalexander https://github.com/moneyadviceservice/frontend/blob/customisable_footer/lib/core/repository/footer/cms.rb#L5

rnalexander:
:+1:   ThanksPaul!

tomas-stefano:
If you dont mind I would like to pull locally and see this in more details.

config/initializers/core.rb
```diff
@@ -30,6 +30,14 @@
...
   Core::Repository::CMS::CmsApi.new
 
+if Feature.active?(:cms_footer)
+  Core::Registry::Repository[:footer] =
+    Core::Repository::Footer::CMS.new
+else
+  Core::Registry::Repository[:footer] =
+    Core::Repository::Footer::Static.new
+end
+
```
tomas-stefano: Is the footer cached? Otherwise it will be one more CMS request for each page request.

app/controllers/application_controller.rb
```diff
@@ -10,6 +10,12 @@ class ApplicationController < ActionController::Base
   include Chat
   include Localisation
 
+  before_action :fetch_footer_content
+  def fetch_footer_content
+    interactor = Core::FooterReader.new('footer')
```
habitullence: Is there any point setting this variable if it's only used once in scope?

paulanthonywilson: Deep question! It's usually done for readability. Fowler calls them "explaining variables" (see Refactoring, p124 "Introduce Explaining Variable").

config/initializers/core.rb
paulanthonywilson: Not cached. Is the extra request a problem?

I'm not at all sure how important this feature is (being able to edit the footer); the only real solution is to drop it if we are worried about the extra request; there isn't enough time to take another approach.

tomas-stefano: About the importance of the feature, I am suspicious to say. But, there is already other place using Cache. It is pretty straight forward to use.

paulanthonywilson: Oh, that might bet worth taking a look at. Can you point us in the right direction?

tomas-stefano: Sure. You can use this class: https://github.com/moneyadviceservice/frontend/blob/master/lib/core/repository/cache.rb

KateWilkinson: @tomas-stefano thanks for this Tomas, I've adjusted to use Cache as suggested

tomas-stefano: Ok. I will take a look.

tomas-stefano:
:+1:

paulanthonywilson:
:+1: Nice footer caching. We should make it clear to the content team / Umair that it can take up to an hour for footer changes to appear

tomas-stefano:
On QA for the guys tested you, since is memory store, you can restart the application process after the changes are made on CMS.

UmairIft:
Thanks @paulanthonywilson , point noted

paulanthonywilson:
Yeah, good point for QA @tomas-stefano

PR #1397, opened 2016-02-25
[DEV APPROVED] CollapsableMobile component extends from Collapsable.

nmiddleweek:
This imports the new CollapsableMobile component from Dough

### DIFF
--- a/app/assets/javascripts/require_config.js.erb
+++ b/app/assets/javascripts/require_config.js.erb
@@ -98,6 +98,7 @@
     utilities: requirejs_path('dough/assets/js/lib/utilities'),
 
     Collapsable: requirejs_path('dough/assets/js/components/Collapsable'),
+    CollapsableMobile: requirejs_path('dough/assets/js/components/CollapsableMobile'),
     DoughBaseComponent: requirejs_path('dough/assets/js/components/DoughBaseComponent'),
     FieldHelpText: requirejs_path('dough/assets/js/components/FieldHelpText'),
     RangeInput: requirejs_path('dough/assets/js/components/RangeInput'),


johnplayer1982:
Nice! :+1:

slaywell:
Great, can this be unit tested?

app/assets/stylesheets/components/common/_collapsable-mobile.scss
```diff
@@ -0,0 +1,26 @@
...
+  .collapsable__trigger-icon {
+    float: right;
+    margin: 10px 20px 0 5px;
+
+    @include respond-to($mq-m) {
+      display: none;
+    }
+  }
+
+  button {
```
slaywell: Can we use BEM here instead of the element? Also, can we remove the nesting and make the class names consistent with the component name?

nmiddleweek: Unfortunately, in this situation we can't use pure BEM, the parent Collapsable Component injects the button.

the ```. collapsable__trigger-icon``` is an element of the nested (Parent) component and I can't flatten that out of the nested class otherwise it'll apply to and therefore break all the other Collapsable components.

slaywell: Cool, I thought that might be the case!

munckymagik:
Is this something that could be shared into Dough? It sounds like it's generic enough?

nmiddleweek:
Dan - yeah, good suggestion. I'll move it over into Dough.

nmiddleweek:
This has been moved to Dough however there are no unit tests at the moment because we're researching how to do Responsive Tests for this comp.

slaywell:
:+1:

munckymagik:
:+1:

rnalexander:
:+1:

PR #1402, opened 2016-03-03
Remove feature flag and unused static footer

paulanthonywilson:
As a tidy up exercise, we now now longer need the static footer. Also
the feature flag for getting the footer from the CMS is no longer
needed; the footer is always got from the CMS.


mconnell:
:+1:

Tr4pSt3R:
:+1:

paulanthonywilson:
Thanks all

PR #1416, opened 2016-03-15
7208 - Add missing logo file for external sites

nmiddleweek:
Since the migration we've lost a logo_en.png file which is being referenced from some external websites and we need to get this fix ASAP as it's a sensitive time of year... Here's an example of where it's missing....

image

There's two parts to fixing this...

1.  Upload to /assets/images/ (This PR)
2. Do some magic in Scripts with the help of Devops so we don't don't redirect to the CDN for this particular image.


slaywell:
:+1:

davidtrussler:
:+1:

PR #1420, opened 2016-03-23
[DEV APPROVED] fix expectation test for Sticky newsletter

Tr4pSt3R:
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1420)

### DIFF
--- a/features/display_newsletter_form.feature
+++ b/features/display_newsletter_form.feature
@@ -15,5 +15,11 @@ Feature: Display sticky newsletter
   @javascript
   Scenario: should see the sticky newsletter footer after scrolling down
     When I scroll to the bottom of the page
+    Then I should see the newsletter form
+
+
+  @javascript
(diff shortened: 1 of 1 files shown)


davidtrussler:
:+1:

jongilbraith:
Right, have now had a proper look at this as I was the last one to work on this.

So, the steps don't match up with the scenario description. I changed it so that it should just test the footer appearing, so the dismiss step should just not be in there.

If you want to actually test the dismissal, then we should add a new scenario for that.

rnalexander:
:+1: Good catch.

Tr4pSt3R:
@jongilbraith could you have another look please?

jongilbraith:
@Tr4pSt3R excellent :smile:  :+1:

rnalexander:
:+1:

Tr4pSt3R:
Thanks @jongilbraith @rnalexander @davidtrussler

PR #1421, opened 2016-03-23
[DEV APPROVED] 7034 - Skip to main link update

johnplayer1982:
## Skip to main content links update for Articles and Category pages

Skip to main link on categories and articles pages are being adjusted to that they jump to the main content.  This change will remove the main id from the entire content and place it on the actual main content of both the article and category pages.

### Tests
The tests check if the container <main> in both article and category pages contain the attribute ``id="main"`` as this is what the accessibility link contains as the anchor ``<a href="#main">``

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1421)

### DIFF
--- a/features/step_definitions/skip_to_main_link_steps.rb
+++ b/features/step_definitions/skip_to_main_link_steps.rb
@@ -1,0 +1,29 @@
+When(/^I visit the homepage$/) do
+  home_page.load
+end
+
+When(/^I view an article$/) do
+  browse_to_article article('en')
+end
+
+When(/^I visit a category page$/) do
+  browse_to_category(category, 'en')
+end
+
+Then(/^the main area is immediately after the header$/) do
+  expected_element = page.find('.t-default-main-content')
+  main_element = page.find('div#main')
+  expect(main_element).to eq expected_element
+end
+
+Then(/^the main area is at the article content$/) do
+  expected_element = page.find('.t-article-main-content')
+  main_element = page.find('main#main')
+  expect(main_element).to eq expected_element
+end
+
+Then(/^the main area is at the categories list$/) do
+  expected_element = page.find('.t-category-main-content')
+  main_element = page.find('main#main')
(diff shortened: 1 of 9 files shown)


features/step_definitions/skip_to_main_link_steps.rb
```diff
@@ -0,0 +1,23 @@
...
+When(/^I view an article$/) do
+  browse_to_article article('en')
+end
+
+When(/^I visit a category page$/) do
+  browse_to_category(category, 'en')
+end
+
+Then(/^I should be able to skip to the main content$/) do
+  expect(page).to have_selector('div#main')
```
jongilbraith: @johnplayer1982 not sure about this mate, these three steps are all just checking that there is a `main#main` element on the page, regardless of where.

What I had in mind when we were chatting before was something along the lines of adding `t-*` selectors to the relevant elements, then using those to see if they were the same elements as `main#main`.

Something along the lines of:

```
expected_element = page.find('.t-default-main-content')
main_element = page.find('main#main')
expect(main_element).to eq expected_element
```

johnplayer1982: Cool thanks @jongilbraith, I'll take a look in the morning

slaywell:
:+1: for the frontend part / accessibility improvements.

jongilbraith:
:+1: though I would just have one final small improvement to suggest with the wording of the steps.

Rather than `I can skip to the categories list`, etc I think it might describe what the test is doing a bit better if you said something like `the main area is at the categories list`.

If the test was actually clicking the link and the end step was doing some sort of test of where the viewport had moved to (which would actually be a pain to do!) then how it is now would suit it well, but given what the test is actually doing then I reckon the alternate wording would be better suited.

jongilbraith:
:+1: perfect :-)

PR #1438, opened 2016-04-13
[WIP] - Grouping gems

tomas-stefano:
We discussed in a meeting about versioning, and this is the start point to group the gem in the right source to avoid conflict. The benefit is that removes the github option for some of the gems (unfortunately not all gems),

All rspec tests pass and cucumber ones too.

---

This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1438)


Gemfile
```diff
@@ -1,5 +1,30 @@
...
+  gem 'cutback_calculator', '~> 0.10'
+  gem 'debt_advice_locator', '~> 2.0'
+  gem 'debt_and_mental_health', '~> 1.0.1'
+  gem 'debt_free_day_calculator', '~> 2.1.0'
+  gem 'debt_test'
+  gem 'decision_trees', '~> 2.0.0'
+  gem 'feedback', '~> 0.3.1'
+  gem 'mortgage_calculator', '~> 1.3.5'
+  gem 'mas-assets'
+  gem 'meta-tags', github: 'moneyadviceservice/meta-tags'
```
tomas-stefano: The meta-tags is not being upload to mas server. It is using a fork, so we should see what should we be done with those (but not on these PR).

Gemfile
```diff
@@ -1,5 +1,30 @@
...
+  gem 'payday_loans_intervention', '~> 1.4'
+  gem 'pensions_calculator', '~> 1.0.0'
+  gem 'quiz'
+  gem 'rio', '= 0.0.3', source: 'http://gems.dev.mas.local'
```
tomas-stefano: Bundler complain about different sources, so I forced and the warnings were removed.

misaka:
I like this change, but as discussed, we need to investigate the changes introduced by the gem updates here.

Gemfile
```diff
@@ -1,5 +1,30 @@
...
+source 'http://gems.dev.mas.local' do
+  gem 'action_plans', '~> 4.0.1'
+  gem 'advice_plans', '~> 3.1.0'
+  gem 'agreements', '~> 2.0.1'
+  gem 'baby_cost_calculator'
+  gem 'budget_planner', '~> 4.0.3'
+  gem 'car_cost_tool', '~> 1.0.4'
+  gem 'contribution_calculator', '~> 0.0.1'
+  gem 'cost_calculator_builder', '~> 0.1.10'
+  gem 'cutback_calculator', '~> 0.10'
```
misaka: Should we be locking this to `~> 0.10.0`?

Gemfile
```diff
@@ -1,5 +1,30 @@
...
+  gem 'debt_advice_locator', '~> 2.0'
```
misaka: Should we be locking this to `~> 2.0.0?`

Gemfile
```diff
@@ -1,5 +1,30 @@
...
+  gem 'debt_and_mental_health', '~> 1.0.1'
+  gem 'debt_free_day_calculator', '~> 2.1.0'
+  gem 'debt_test'
+  gem 'decision_trees', '~> 2.0.0'
+  gem 'feedback', '~> 0.3.1'
+  gem 'mortgage_calculator', '~> 1.3.5'
+  gem 'mas-assets'
+  gem 'meta-tags', github: 'moneyadviceservice/meta-tags'
+  gem 'payday_loans_intervention', '~> 1.4'
```
misaka: Should we be locking this to `~> 1.4.0?`

Gemfile
tomas-stefano: Yes, we should.

tomas-stefano:
@misaka I added the user story here: https://moneyadviceservice.tpondemand.com/entity/7261

PR #1439, opened 2016-04-13
[DEV APPROVED] 7233 signup article end copy

nmiddleweek:
After the A/B Test proved an increase in signup's in having a 'generic' paragraph at the bottom of each article, we've decided to include this as a permanent fixture. This screen shot shows the change visually,

image

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1439)


app/views/articles/show.html.erb
```diff
@@ -30,13 +30,16 @@
       <%= heading_tag article.title %>
 
       <% unless @newsletter_excluded %>
-        <%= render 'shared/news_signup_test',
-                   ab_test_render: true %>
+        <%= render 'shared/news_signup_test', ab_test_render: true %>
         <%= render 'shared/news_signup_inpage', ab_test_render: false %>
```
jiannis: Do we still need the _test_ version of this partial now that the A/B test is over?

app/views/articles/show.html.erb
```diff
@@ -30,13 +30,16 @@
...
       <% end %>
 
       <%= article.content %>
 
+      <% unless @newsletter_excluded %>
+        <%= render 'inline_newsletter_sign_up' %>
```
jiannis: Maybe we should rename this partial to `inline_newsletter_signup` to be consistent with the naming of the shared partials above?

app/views/articles/show.html.erb
nmiddleweek: This isn't part of this story so the removal of ```<%= render 'shared/news_signup_inpage', ab_test_render: false %>``` will be part of another ticket.

config/locales/en.yml
```diff
@@ -315,6 +315,11 @@ en:
       popular_links_title: Most read
       related_links_title: Related guides
 
+    inline_sign_up:
```
jiannis: Maybe we should group this `inline_sign_up` with [`newsletter_subscriptions`](https://github.com/moneyadviceservice/frontend/blob/7233-signup-article-end-copy/config/locales/en.yml#L385) and rename it to `inline_signup`?

_Side Note_: There is also a [`signup_inpage`](https://github.com/moneyadviceservice/frontend/blob/7233-signup-article-end-copy/config/locales/en.yml#L406) with similar content which is a bit of a smell :dash:

app/views/articles/show.html.erb
nmiddleweek: Good spot.

config/locales/en.yml
nmiddleweek: Yes, good point however this paragraph is very article specific and just above the sharing icons so I think it's better suited above the article.sharing locale section.

n.b. I'm sure the whole locale files file could be refactored to be better organised site-wide but I don't want to open a can of worms and delay getting this released. Perhaps we can create a ticket to organise this?

app/views/articles/show.html.erb
nmiddleweek: This is updated

jiannis:
@nmiddleweek agree that cleaning up stale code from the A/B test is not part of this story but we should do some spring cleaning as another card. 

The fact that we have two partials - [`news_signup_test`](https://github.com/moneyadviceservice/frontend/blob/723bb538671854b79e91ab43ecd75c8cf472a2a0/app/views/shared/_news_signup_test.html.erb) and [`news_signup_inpage`](https://github.com/moneyadviceservice/frontend/blob/723bb538671854b79e91ab43ecd75c8cf472a2a0/app/views/shared/_news_signup_inpage.html.erb)- as well as two implementations - [`newsletter_exclusions`](https://github.com/moneyadviceservice/frontend/blob/723bb538671854b79e91ab43ecd75c8cf472a2a0/lib/newsletter_exclusions.rb) and [`sticky_newsletter_visibility`](https://github.com/moneyadviceservice/frontend/blob/723bb538671854b79e91ab43ecd75c8cf472a2a0/app/helpers/sticky_newsletter_visibility.rb) - to prevent the newsletter signup from appearing on certain pages is a bit funky :dancer: :circus_tent:

Happy to help sort this out if it's OK with Karine

jiannis:
Not super happy. We can do better. Good enough for now :+1:

The following two cards are about cleaning this up:

- [Clean Up A/B Newsletter Signup Experiment](https://moneyadviceservice.tpondemand.com/entity/7262)
- [DRY Logic to Prevent Newsletter Sign Up on Sensitive Pages](https://moneyadviceservice.tpondemand.com/entity/7263)

tomas-stefano:
:+1: (One thing in mind is that we should plan to remove all shared views and transform into resourceful views - I will plan a lighting talk on the subject)

nmiddleweek:
Closing to open QA PR

nmiddleweek:
Opening again now it's merged into QA

PR #1447, opened 2016-04-28
[DEV APPROVED] TP: 7225, Comment: Adds Back To Top functionality to articles

davidtrussler:
This adds a component to Dough that adds a link for smaller devices to return the user to the top of the page. This is activated when the user scrolls down the content by a certain amount.

Depends on PR#268 in Dough: https://github.com/moneyadviceservice/dough/pull/268

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1447)


app/assets/stylesheets/layout/page_specific/_article_page.scss
```diff
@@ -109,3 +109,65 @@
...
+  transform: translateY(-50%);
+  fill: $color-white;
+}
+
+.article-3col-main__btt-link:hover,
+.article-3col-main__btt-link:focus {
+  text-decoration: underline;
+}
+
+.footer--btt_link {
```
slaywell: This doesn't really fit with our BEM conventions. I'd suggest that we put the 'btt-link' in a separate file and then have 'footer' or 'article' as the modifier?

app/views/articles/show.html.erb
```diff
@@ -24,7 +24,11 @@
   <% end %>
 <% end %>
 
-<div class="l-article-3col-main editorial">
+<div
+  class="l-article-3col-main editorial"
+  data-dough-component="BackToTop"
+  data-dough-back-to-top-config='{"i18nStrings": {"title":"<%= t('articles.nav.back_to_top') %>"}}'
```
slaywell: Nice!

app/assets/stylesheets/layout/page_specific/_article_page.scss
davidtrussler: Not sure - it's actually modifying the footer when the link is present (to give that bit of extra space at the bottom of the page). It probably would be better in the footer styles though.

slaywell:
Definitely better now that the CSS is in its own file, but it's not following our BEM naming conventions. If the file is called `_back_to_top` then all the classes within should begin with `. _back_to_top`and either be elements (e.g. `._back_to_top__link`) or modifier classes.

Also, the color contrast of the button is too low (Contrast Ratio: 3.01:1, and it's not 'large text') so we should ask if we can use a different tone for this.

slaywell:
👍

johnplayer1982:
Thats awesome! :+1:

slaywell:
👍 still applies!

PR #1453, opened 2016-05-13
TP: 7311, Comment: Bumps version for Timelines

davidtrussler:
Required for PR#63 in Timelines (https://github.com/moneyadviceservice/timelines/pull/63)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1453)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,7 +63,7 @@ gem 'pensions_calculator', '~> 1.1.1'
 gem 'quiz', git: 'git@github.com:moneyadviceservice/quiz.git'
 gem 'rio', '= 1.3.0', source: 'http://gems.dev.mas.local'
 gem 'savings_calculator', '~> 1.2.0'
-gem 'timelines', '>= 1.2.0', git: 'git@github.com:moneyadviceservice/timelines.git'
+gem 'timelines', '~> 1.2.0', git: 'git@github.com:moneyadviceservice/timelines.git'
 
 # 1.0.2 has breaking changes as it adds japanese and turkish locales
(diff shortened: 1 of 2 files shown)


slaywell:
Do we want to switch to the pessimistic operator `~>` rather than `>=` based on the semantic versioning / dependencies discussion we had before?

davidtrussler:
@slaywell Probably so, more or less everything else is using that. Would that mean there is no need to bump for the "patch" iteration (i.e. ~> 1.2.0)?

slaywell:
I *think* so, but I might be wrong. If you bundle without doing this version bump, does it pull your latest changes? I think we should change it though.

jiannis:
@davidtrussler You are quite right, `~> x.y.z` means _"equal to or greater than the last digit_". Specifying `~> 1.2.0` in the `Gemfile` and running `bundle` would pick up versions _"equal to or greater than 1.2.0 but less than 1.3.0"_, i.e. the latest patch 1.2.x version. You'd still run `bundle` to update `Gemfile.lock` and commit that.

jiannis:
So `~> 1.2.0` would be OK too since [timelines provides a gemspec with versions](https://github.com/moneyadviceservice/timelines/blob/master/lib/timelines/version.rb)

davidtrussler:
@jiannis So is the best practice here to use ~> 1.2.0 in the Gemfile?

johnplayer1982:
:+1:

jiannis:
Squash and 👍

slaywell:
👍

PR #1454, opened 2016-05-18
[DEV APPROVED] 7347 articles remove newsletter sign up

davidtrussler:
This PR removes the two Newsletter signup sections from article pages only. 

image

image

It removes:
- references to the partial that contains the in-article view (but not the partial itself which is still needed on the Contact Us page) and the partial that contains the inline view
- the JS component which controls the display of the in-article view
- tests associated with that component
- the partial that contains the view for the inline sign-up section
- YAML text for the inline sign-up article (only in the English file since the signup doesn't appear in Welsh)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1454)


slaywell:
Looks good, am I right that we also can't remove any styles for this until we remove the component from the Contact Us page?

davidtrussler:
@slaywell Yes, that's correct. I have added notes about that into the ticket for the Contact Us page (https://moneyadviceservice.tpondemand.com/entity/7314), along with additional stuff that needs to go from the application controller, that is still needed until that point.

slaywell:
Sounds good! 👍

johnplayer1982:
Very thorough :+1:

PR #1463, opened 2016-06-06
[DEV APPROVED] 7401 Frontend SCSS refactor into Yeast

slaywell:
This is the corresponding PR in Frontend to [this Yeast refactoring PR](https://github.com/moneyadviceservice/yeast/pull/15).

Interestingly I can't get glob syntax working from a Bower component. So whereas before we had;

```@import 'components/dough_theme/*/_all.scss';```

This does not work:
```@import 'yeast/assets/components/dough_theme/*/_all.scss';``` - causes a file not found or unreadable SCSS error.

I have had to work around this by adding a new file in Yeast that imports all the Dough Theme components, but if anyone has any ideas on why it doesn't work I would be interested!

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1463)


davidtrussler:
👍

munckymagik:
👍

PR #1473, opened 2016-06-14
Change slug of article used for company list on the debt management page

jongilbraith:
This page is unique in that it makes a call to the CMS for an article and then inserts that amongst other content in the template.

[Long story short](https://moneyadviceservice.tpondemand.com/tp2/entity/7422), we need to change the slug of the article in question. The page has been duplicated in the production CMS in an article with the new slug, so this updates frontend to use that new slug.

Once deployed, we'll then unpublish the old article.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1473)


misaka:
:+1:

pmcostello:
👍 looks good to me

rnalexander:
:+1: 🚢

PR #1468, opened 2016-06-10
[DEV APPROVED]Remove contact-us static page and accoutrements

misaka:
The goal is to allow contact-us to be handled by the CMS. The issue is
that this page is a lynch-pin for the machinery that serves out static
pages. Without it tests fail as there's nothing left to serve out, so
this change involves removing the static page machinery and the related
tests.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1468)

PR #1474, opened 2016-06-15
[DEV APPROVED] Remove contact-us static page and accoutrements

misaka:
The goal is to allow contact-us to be handled by the CMS. The issue is
that this page is a lynch-pin for the machinery that serves out static
pages. Without it tests fail as there's nothing left to serve out, so
this change involves removing the static page machinery and the related
tests.

(re-opened PR because I'm an idiot that likes to merge into master before QA is done ... this PR has been previously approved as #1468)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1474)


Tr4pSt3R:
👍 FWIW

PR #1477, opened 2016-06-21
7416 - newsletter subscription cleanup

jongilbraith:
After removing the ability to opt into the newsletter on signup and in the user settings, it was realised we no longer would need all the backend code in `lib/core` that interacts with the CRM system when changing their subscription status.

Additionally we realised there was loads of left over functionality which was overlooked when we removed things like calls to action (a javascript back end for checking and changing subscription status), so I've removed those.

All that's left now is `newsletter_subscription` as an attribute of the user which I've left in place, since we're leaving that in place in the user model and preserving the historical data.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1477)


rnalexander:
Awesome job catching the extra orphaned code!  :+1:

spec/spec_helper.rb
```diff
@@ -48,11 +48,12 @@
...
-               'raw_web_chat_additional_three', 'raw_web_chat_small_print', 'raw_contact_heading',
-               'raw_contact_introduction', 'raw_contact_phone_number', 'raw_contact_additional_one',
-               'raw_contact_additional_two', 'raw_contact_additional_three', 'raw_contact_small_print',
-               'raw_newsletter_heading', 'raw_newsletter_introduction'].map do | identifier |
+    blocks = [
+      'raw_web_chat_heading', 'raw_web_chat_additional_one', 'raw_web_chat_additional_two',
+      'raw_web_chat_additional_three', 'raw_web_chat_small_print', 'raw_contact_heading',
+      'raw_contact_introduction', 'raw_contact_phone_number', 'raw_contact_additional_one',
+      'raw_contact_additional_two', 'raw_contact_additional_three', 'raw_contact_small_print'
+    ].map do | identifier |
```
misaka: This kind of fix makes me happy. 😄

misaka:
:+1: Looks good to me. Is this something that'll go through a quick QA or straight to staging?

johnplayer1982:
Nice!, bye bye newsletter :+1:

jongilbraith:
@misaka was going to say it might be good to pop this onto qa to check the crm integration definitely hasn't broken anything, but not sure if we have all of that on that environment?

misaka:
@jongilbraith You're likely right about us not having it enabled for that environment. Should check with @pmcostello, he was playing around with that recently iirc.

pmcostello:
Nice job! we can use the CRM test environment currently plugged into UAT to test the changes in this PR
👍

jongilbraith:
Great, thanks @pmcostello, I'll get it onto UAT when that's free

PR #1486, opened 2016-07-15
[DEV APPROVED] 7444 - SVG title attributes

johnplayer1982:
## Adding title attributes to SVG's

This ticket adds title attributes to the footer social SVG's.  This will also need to apply to other projects - Blog, RIO and any other that does not use the frontend footer

screen shot 2016-07-15 at 15 04 01

Please note: There is a related ticket in dough to apply the same changes to the social sharing icons:
https://github.com/moneyadviceservice/dough/pull/272

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1486)

### DIFF
--- a/app/views/shared/_footer_primary.html.erb
+++ b/app/views/shared/_footer_primary.html.erb
@@ -6,7 +6,8 @@
         <li class="l-footer-primary__list--left">
           <div class="footer-primary__list-item">
             <a class="button facebook-link t-facebook-link" lang="en" href="https://www.facebook.com/MoneyAdviceService?ref=mas" target="_blank">
-              <svg xmlns="http://www.w3.org/2000/svg" class="svg-icon svg-icon--facebook">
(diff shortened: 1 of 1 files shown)


slaywell:
I might be wrong here, but I thought `role="presentation"` was for when the image is entirely decorative and should be ignored by screen readers (so each SVG either needs a title _or_  role="presentation")?

johnplayer1982:
Yes I think you're right, These should just have titles, as they aren't just for presentation

johnplayer1982:
@slaywell The update has been commited

slaywell:
Tip top 👍

davidtrussler:
Might need to add the email one as well - this is used on the articles pages, which doesn't use YouTube.

johnplayer1982:
Thanks @davidtrussler The article social sharing icons have been updated and are in a seperate PR in dough:

https://github.com/moneyadviceservice/dough/pull/272

davidtrussler:
👍

PR #1488, opened 2016-07-18
[DEV APPROVED] - 7440 - Null image ALT attributes

johnplayer1982:
## The issue

The DAC report highlights that images on the following page contain descriptive alt attributes:
https://www.moneyadviceservice.org.uk/en/pensions-and-retirement/pension-savings-timeline
here https://www.moneyadviceservice.org.uk/en/pensions-and-retirement

However due to accessibility requirements it is necessary to have a null alt attribute.

## The solution

The solution was to remove the alt attributes for the images from the yaml files (en|cy) and update the rails image_tag helper to use a null alt attribute like so:

```
<%= image_tag "retirement_concepts/in-retirement@2x.png", alt: '' %>
```
Which results in the output:
```
<img alt="" src="/assets/retirement_concepts/in-retirement@2x.png">
```

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1488)

### DIFF
--- a/app/views/retirements/pension_savings/_five_years_out_section.html.erb
+++ b/app/views/retirements/pension_savings/_five_years_out_section.html.erb
@@ -7,7 +7,7 @@
             <%= t('pension_savings_timeline.thinking_about_retirement_five_years_out.title') %>
         </h2>
         <div class="graphic-info-box__illustration">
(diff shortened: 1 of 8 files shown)


slaywell:
👍 Checked locally, looks good!

jongilbraith:
👍

PR #1496, opened 2016-07-26
[DEV APPROVED] 7480 - Removing want to talk promo panel

johnplayer1982:
## Removing want to talk hidden panel

There is a 'money worries' promo slot hidden in the right hand column. It contains invalid html - an h2 within button tag. This is not rendered on screen - the whole thing is hidden with CSS.
 
It looks like this is old code that needs to be removed.

### Solution includes:

- Removing partial
- Removing JS
- Removing CSS
- Removing associated tests

I have also checked CMS and Dough for any associated content, this is unique to frontend so no further action required on another repo.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1496)


tomas-stefano:
👍

davidtrussler:
👍

PR #1497, opened 2016-07-26
[DEV APPROVED]7430 populate content hub from yaml

davidtrussler:
This integrates some real content into the Content Hub template we produced a while back, and stores it in a YML file (at this stage we only have an English one, with Welsh on the way). 

I will be deploying this to QA shortly so the product  owners can test it properly and marked it as WIP as there will be more work to do on it (especially when the Welsh language content is added). 

In the meantime it wold be useful to discuss the code.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1497)


app/views/content_hub/blocks/_double.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="l-content-hub__module l-content-hub__module--hdr-con-lnk-img">
+  <div class="l-content-hub__module-inner">
+    <%= heading_tag block[:header], class: 'content-hub__module-title heading-medium', level: 1 %>
```
slaywell: Don't think this should be level: 1 - at the moment every header on the page is coming out as an `<h1>`

app/assets/stylesheets/layout/page_specific/_content_hub.scss
```diff
@@ -0,0 +1,234 @@
+// Content Hub Variables --
+// Colours
+$content-hub-title-color: $color-black;
+$content-hub-bgcolor: $color-grey-eight;
```
slaywell: Minor point but can we use `$color-grey-pale` - this will be the same colour as the breadcrumb header. We have waaaay too many different tones to pick from!

app/views/content_hub/blocks/_list_with_cta.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <div class="l-content-hub__module-inner">
+    <%= heading_tag block[:header], class: 'content-hub__module-title heading-medium', level: 1 %>
+    <ul class="content-hub__module-links unstyled-list">
+      <% block[:links].each do |link| %>
+        <li><a href="<%= link[:url] %>"><%= link[:text] %></a></li>
+      <% end %>
+    </ul>
+    <a href="#" class="content-hub__module-cta">
+      <%= block[:cta] %>
+      <svg xmlns="http://www.w3.org/2000/svg" class="svg-icon content-hub__module-cta-icon">
```
slaywell: Could you add `role="presentation"` so screen readers ignore this svg?

app/assets/stylesheets/layout/page_specific/_content_hub.scss
```diff
@@ -0,0 +1,234 @@
...
+$content-hub-module-reverseBGcolor: $color-black;
+
+// Widths
+$content-hub-page-width: gridsystem-width(12) - $gutter-width; // unit: px
+$content-hub-modules-gutter: $baseline-unit * $base-font-size * 1px / ($baseline-unit * 0 + 1); // unit: px
+$content-hub-horizontal-space: $content-hub-modules-gutter * 3; // unit: px
+$content-hub-width-one-module: px((($content-hub-page-width) - ($content-hub-modules-gutter * 2)) / 3px); // units: em
+$content-hub-width-two-module: ($content-hub-width-one-module * 2) + px($content-hub-modules-gutter / ($content-hub-modules-gutter * 0 + 1)); // units: em
+$content-hub_width_three-module: px($content-hub-page-width / ($content-hub-page-width * 0 + 1)); // units: em
+$content-hub-module-margin-width: (px(($content-hub-horizontal-space) / ($content-hub-horizontal-space * 0 + 1)) / $content-hub-width-one-module) * 100%; // unit: %
```
slaywell: This is getting complex... Maybe it's time for a new grid system that's more flexible for gutter-less grids! OK for this page though I think.

johnplayer1982:
All looks great, apart from a small observation:

The contact panels have a top border, I don't think it needs it as the separation between content is done with the background colour difference:

screen shot 2016-08-16 at 11 12 49

davidtrussler:
@johnplayer1982 I agree on your contact panels comment though it might be more trouble than it's worth to remove it since it's part of the standard footer.

johnplayer1982:
Ah right i see, tricky! and your right it would mean a new layout! in that case it gets a :+1: from me, this is really good!

rnalexander:
👍 :shipit:

PR #1503, opened 2016-08-04
[DEV APPROVED] Add pre-build script

misaka:
Move common functions from test.sh and build.sh into pre-build.sh so as
to reduce the chances of race conditions between those two. This change
requires a change to the Go pipeline to run pre-build before the
test_and_build job.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1503)


tomas-stefano:
Pre-build: I remember you! #Go 👍

jongilbraith:
👍

misaka:
I should've mentioned yesterday, before this gets merged to master we need to:
- change the `parallel_test_and_rpm_package_new` pipeline to remove the precompiled assets
- change the `responsive_commit_preview` pipeline to use the `parallel_test_and_rpm_package_new` template, check if any variables need to be set for the `preview` environment.

misaka:
Turns out ... this is not what we need. We're removing the pre-build script from the blog too (see https://github.com/moneyadviceservice/publify/pull/198/commits/37daca2d6a4e5447525475717148250ca003b31e) and going to make bower operations insulated to each agent running them.

PR #1513, opened 2016-08-23
[DEV APPROVED]TP: 7520, Comment: Updates typeahead functionality to exclude tab key

davidtrussler:
This PR adds typeahead.jquery.js as a local asset (was previously loaded as a Bower component) and makes some changes (shown below) to it to prevent the current behaviour where the tab key has the same behaviour as the enter key (fires a search from a highlighted search suggestion). 

image

This was proving an accessibility issue: the tab key would be expected to move the user to the next link (in this case the clear button) rather than fire a search.

Also updates some associated files necessary for this change.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1513)

### DIFF
--- a/app/assets/javascripts/require_config.js.erb
+++ b/app/assets/javascripts/require_config.js.erb
@@ -4,7 +4,7 @@
 //= depend_on_asset jquery/dist/jquery
 //= depend_on_asset jquery-waypoints/waypoints
 //= depend_on_asset jquery-ujs/src/rails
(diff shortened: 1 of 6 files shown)


slaywell:
Please could you point out where in the vendor JS file you've removed the tab handler function?

davidtrussler:
@slaywell I've added a screenshot of a git diff of the original and changed files.

slaywell:
👍

rnalexander:
:+1: :shipit:

PR #1515, opened 2016-08-24
[DEV APPROVED]7568 - Adding svg-icon--arrow-left to SVG partial

slaywell:
This icon is useful and can be reused, so we are adding it to the main SVG partial that appears on every page on the site (the arrow-right is already in there).

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1515)


davidtrussler:
👍

rnalexander:
:+1: :shipit:

PR #1518, opened 2016-08-25
[DEV APPROVED] - TP: 7569, Comment: Updates element that has focus after search

davidtrussler:
This addresses an accessibility issue whereby the focus after a search returns to the top of the page rather than to the first result or, where there are no results, to the input field.

It is written as a Dough component so that it can be shared with the MAS blog to deal with the same issue. Depends on PR 275 in Dough (https://github.com/moneyadviceservice/dough/pull/275)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1518)


slaywell:
@davidtrussler I think this should go back to [WIP] status whilst the corresponding Dough PR is being worked on?

davidtrussler:
@slaywell Yes, thanks for changing that. Also any idea if we need to be concerned about the failing check?

slaywell:
I would ask @Tr4pSt3R about Hakiri - I don't think it's related to this PR specifically

johnplayer1982:
Looks good, same implementation as the publify one :+1:

slaywell:
👍  💯

johnplayer1982:
:+1: for the additional commit

slaywell:
Another 👍  from me

PR #1521, opened 2016-09-02
TP: 7598, Comment: Adds Welsh YML file for Content Hub

davidtrussler:
Adds the Welsh version to the Content Hub page. The English version is on Production here: https://www.moneyadviceservice.org.uk/en/hub/bank-accounts and the YML for that is here: https://github.com/moneyadviceservice/frontend/blob/master/config/locales/content_hub/bank_accounts.en.yml

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1521)

### DIFF
--- a/app/views/content_hub/show.html.erb
+++ b/app/views/content_hub/show.html.erb
@@ -37,3 +37,10 @@
     </div>
   </div>
 </div>
+
+<% content_for(:alternate_link) do %>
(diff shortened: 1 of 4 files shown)


jongilbraith:
LGTM 👍

jriga:
👍

jongilbraith:
@davidtrussler just had a thought about this, does the welsh link correctly plumb in the right slug? I'm thinking probably not, so possibly we should deal with that before rolling this out.

slaywell:
Great! 👍

PR #1527, opened 2016-09-12
[DEV APPROVED] Adding informizely tag to our JS structure

slaywell:
The aim of this PR is to move the Informizely snippet out of Google Tag Manager and into our own site code. We needed to write some custom code to set and read cookies that affect the Informizely exit poll, so rather than writing this code in GTM it is better to have it under version control and loaded asynchronously using RequireJS.

## Why?
We are adding display logic to the exit poll, so that:
- If a user closes the poll, it does not appear again for 90 days
- If it is displayed three times and the user ignores it, it does not appear again for 14 days

## informizely_tag.js

This file contains the Informizely snippet (slightly altered to work using RequireJS), and when we load their snippet we pass it a custom event handler function `informizelyTag.handleIzEvent` which does two things:
- Increments a 'counter' cookie
- Listen to the events emitted by the Informizely code, and if the survey is completed, set the 90 day cookie

This file also contains helper functions to:
- Read a cookie
- Set the two Informizely cookies

[PR 729 in Scripts adds the env variable](https://github.com/moneyadviceservice/scripts/pull/729)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1527)


davidtrussler:
👍 Looks good. I've pulled this locally but it's not possible to test it fully in that environment.

johnplayer1982:
:+1:

pmcostello:
👍

PR #1531, opened 2016-09-14
[WIP] Adding informizely to Asset pipeline and requireJS

slaywell:
To include the new JavaScript file added in #1527 this also needs to be added to the asset pipeline config as well as the dependency in the requireJs file.

Quite frustrating that it works locally without doing this!

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1531)


pmcostello:
👍

PR #1534, opened 2016-09-15
7510 - Adding path=/ to the Informizely cookie code

slaywell:
The cookies that the Informizely tag code sets need to have 'path = /' otherwise duplicates are getting set across different pages on the site when the code is ran from different locations.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1534)


davidtrussler:
👍

tomas-stefano:
👍

PR #1538, opened 2016-09-19
7510 - Revert all Informizely JS code

slaywell:
The code that we added to conditionally fire / hide the Informizely survey is not working consistently on our Staging environment, so this PR is removing it.

We have found that setting cookies with JS is failing to work on an intermittent basis in Chrome. Rather than debug this further and further, the user benefit that this feature would bring is not worth the risk of it failing.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1538)


johnplayer1982:
:+1:

jongilbraith:
👍

tomas-stefano:
👍

PR #1539, opened 2016-09-19
Avoid to install wrong gems and remove warnings

tomas-stefano:
When I pulled the frontend master I received a warning:

```
Warning: the gem 'quiz' was found in multiple sources.
Installed from: http://gems.dev.mas.local/
Also found in:
  * https://rubygems.org/
You should add a source requirement to restrict this gem to your preferred source.
For example:
    gem 'quiz', :source => 'http://gems.dev.mas.local/'
Then uninstall the gem 'quiz' (or delete all bundled gems) and then install again.
```

So I address the solution on this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1539)


jongilbraith:
And that would explain why that (and the timelines) gem was previously pulling directly from the git repo, which I changed in dac5c6ce5ca99e2eb8517dd90a475a0819227775.

Probably good to do the same thing for the timelines gem.

jongilbraith:
OK, slow down and read things properly, Jon. Not a problem for timelines (saw a public gem called "timeliness" existed, and not "timelines").

jongilbraith:
And so, 👍

Tr4pSt3R:
👍

tomas-stefano:
@jongilbraith But will be good to we wrap up all the MAS gems in the source. I will merge this and we can address this later.

PR #1552, opened 2016-10-07
[DEV APPROVED] 7666 - Search form visual display bug fix

johnplayer1982:
## The problem

There is a current display issue on the production site, where after performing a search on the search page (https://www.moneyadviceservice.org.uk/en/search) for a brief moment the results will be tucked underneath the search input and button section.

screen shot 2016-10-07 at 09 29 15

## The solution

The search form now has a minumum height to ensure that the results begin after the search form element, ``@extend %clearfix;`` was unsuccessful as the typeahead component relies on absolute positioning which rendered the clearfix unable to help in this case.  There is also a conditional which adds a modifier class to the in page search form as this only applies to the form on that page, not the one in the header.

screen shot 2016-10-07 at 09 53 45

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1552)

### DIFF
--- a/app/assets/stylesheets/components/common/_search.scss
+++ b/app/assets/stylesheets/components/common/_search.scss
@@ -46,6 +46,10 @@
   }
 }
 
+.search--inpage {
+  min-height: $baseline-unit*9;
+}
+
 .search__submit {
   @extend .unstyled-button;
   position: absolute;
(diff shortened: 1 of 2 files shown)


slaywell:
👍

slaywell:
👍

jongilbraith:
👍

PR #1553, opened 2016-10-07
[WIP] Secure dynamically rendered campaign pages

Tr4pSt3R:
Our application security monitor has thrown up this issue concerning rendering dynamic paths without being mindful of code injection. This should fix that issue by removing any unintended HTML code.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1553)


app/controllers/campaigns_controller.rb
```diff
@@ -7,12 +7,13 @@ def display_skip_to_main_navigation?
     false
   end
 
+  ::CampaignPage::Campaign::WHITELIST.each do |campaign_page|
+    method_name = campaign_page.underscore
+    class_eval "def #{method_name}; end"
```
tomas-stefano: I read the PR code but I didn't identify the reason of creating a method based on the campaign name and use class eval. Can you explain to me th reason of it, please?

Tr4pSt3R: Hi @tomas-stefano the reason for creating dynamic methods for all campaign paths are as follows:
  - these dynamic methods are non-RESTful actions for rendering the relevant template of the same name. For example `/en/campaigns/csa` will render `app/views/campaigns/csa.html.erb`.
  - the alternative implementation was to direct all campaign pages through the show action and apply logic on the params[:id]
  - the reason stated above was the security risk flagged by _Hakiri_

tomas-stefano: Can these methods be private then? And other thing, I would use define_method instead of class_eval. More details here: http://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html

In case you've doubts about to use define_method:

```ruby
2.1.2 :001 > FOO = %w(a b c)
 => ["a", "b", "c"]
2.1.2 :002 > FOO.each do |foo|
2.1.2 :003 >     end
 => ["a", "b", "c"]
2.1.2 :004 > FOO.each do |foo|
2.1.2 :005 >     define_method "#{foo}" do
2.1.2 :006 >       puts "#{foo}"
2.1.2 :007?>     end
2.1.2 :008?>   end
 => ["a", "b", "c"]
2.1.2 :009 > a
a
 => nil
2.1.2 :010 > b
b
 => nil
2.1.2 :011 > c
c
 => nil
```

Tr4pSt3R:
This implementation is unnecessarily complex, considering the problem we are trying to solve.

I have manually marked it as being false positive on Hakiri. This is safe to do because we have protective mechanisms within the app to secure it from file intrusion.  I will close this for now.

Thanks @pmcostello @tomas-stefano @jriga @jongilbraith

PR #1556, opened 2016-10-10
[DEV APPROVED] 7671 Adding new window label to social links in footer

johnplayer1982:
The social media buttons on the foot of the main site are SVGs, which, when clicked open the relevant social media page in a new window.

If users are not made aware that this will open in a new window then they will become disoriented and unable to use the browser's back button to get back to their page.
 
This has been solved in this PR for some readers by adding 
 
<span class="visually-hidden">Facebook - opens in a new window</span>
 to each of the social media links.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1556)


slaywell:
👍

jongilbraith:
👍

PR #1561, opened 2016-10-13
[DEV APPROVED] 7672 - Adding PDF download icon

johnplayer1982:
The video transcript on this Retirement page is a PDF download, that opens in the browser window. This causes problems for many users, who are subsequently unable to navigate back to the page, and are given no warning that this will happen. 

This changes the functionality to download the file and adds a PDF icon to any link with a download attribute.

| English | Welsh |
|---------|-------|
|screen shot 2016-10-13 at 14 11 35|screen shot 2016-10-13 at 14 11 40|

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1561)

### DIFF
--- a/app/assets/stylesheets/components/common/_download_link.scss
+++ b/app/assets/stylesheets/components/common/_download_link.scss
@@ -1,0 +1,11 @@
+// For links that contain download attribute
+
+.download-link--pdf {
+  &:after {
+    @extend %icon;
+    @extend .icon--pdf;
+    content: '';
+    margin-left: 6px;
+    display: inline-block;
+  }
+}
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/components/common/_download_link.scss
```diff
@@ -0,0 +1,11 @@
+// For links that contain download attribute
+
+a[download] {
+  &:after {
+    @extend %icon;
+    @extend .icon--pdf;
```
slaywell: What if it's a download but not a PDF? It would be good to see the code that generates the markup (or the markup itself). Can we add a class as well as adding the download attribute?

johnplayer1982: Very good point! Classes are a good solution for this, ill take a look, thanks @slaywell

app/assets/stylesheets/components/common/_download_link.scss
```diff
@@ -0,0 +1,15 @@
+// For links that contain download attribute
+
+.download-link {
+
+}
```
slaywell: I don't think this empty declaration is a requirement of BEM - unless you're planning on adding any default block styles to `.download-link` I think this should be removed?

slaywell:
👍

jongilbraith:
👍

PR #1570, opened 2016-10-19
[WIP] TP: 7682, Comment: Adds mail-to section for AB Test

davidtrussler:
Adds some basic markup and CSS to codebase for Optimizely to target in an AB Test. 

Although deploying changes for this purpose is undesirable this is necessary in this instance because (as I understand it) we are adding elements to the generic article page and this cannot be changed via Optimizely at this level, but only at the level of individual articles.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1570)

### DIFF
--- a/app/assets/stylesheets/layout/page_specific/_article_page.scss
+++ b/app/assets/stylesheets/layout/page_specific/_article_page.scss
@@ -109,3 +109,7 @@
     @include column-shift(3);
   }
 }
+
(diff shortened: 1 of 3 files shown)


davidtrussler:
After discussion with @slaywell, closing this PR in order to implement this all through Optimizely.

PR #1576, opened 2016-10-24
[WIP]Use a working cms api end point for development purposes

jongilbraith:
The current url given no longer serves via http and can't be used over https due to the self signed ssl certs, so just use the production cms instead.

---

This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1576)


tomas-stefano:
Since we're introducing POST and PATCH requests to CMS, I'm not sure about this. Can we use localhist instead and talk to everyone do the same?

jongilbraith:
@tomas-stefano good point.

In which case, maybe then we would be better not having any real environment variables checked in at all with the expected setup to be for people to have their own gitignored `.env.*` files.

PR #1577, opened 2016-10-27
[DEV APPROVED] Adds Support for Google AMP

shaley91:
# Google AMP
## What is it?

AMP is a way to build web pages for static content that render fast. To put it simply, AMP is basically a stripped down HTML page which restricts certain HTML, CSS, and Javascript. It’s designed to be used for articles, not across entire sites.
## AMP Requirements:
- Allows only asynchronous scripts
- All CSS must be inlined
- Styles are limited to 50KB
- External resources like images must state their size in HTML
- All Javascript must be out of critical path
- No user written Javascript is allowed, only AMP supported
- Fonts must be loaded with a link tag or a CSS @font-face rule
# Changes in this PR
- We create an `/amp` sub-route for articles that returns a AMP compliant view. 
- We have created a separate Amp Articles controller 
  - This Subclasses `ActionController::Base` rather than `ApplicationController` because we need to limit everything that gets pulled into the view.
- We created a very basic view that only has AMP compliant HTML as well as various other AMP requirements (i.e. importing the AMP JS lib)
- We created a `amp_styles.scss` file.  This includes very little styles because AMP requires that styling is kept bellow 50Kb
- We added a AmpArticle Decorator that replaces video `iframes` with an `amp-youtube` tag that is AMP compliant
- We link AMP Article views with Articles views and vice versa with canonical link tags.

screenshot 2016-10-27 11 11 27

---

This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1577)


rnalexander:
Code Climate raised 5 issues all of which are minor duplications which I believe would make the code more confusing to extract out.  The Hakiri failures are based on cross site scripting which is to be expected when pulling in javascript from another source are I believe we're required to do as part of the AMP spec (although that being said we are trusting google not to mess with us by changing the javascript we load from them.)

TL;DR:  Looks good, I for one welcome our new google overlords.

:+1:

shaley91:
Thanks @rnalexander.  I don't think I have access to code climate.

There are still a few changes that are required (including fixing a few cucumber tests)

rnalexander:
Will the next changes be coming in another PR?

shaley91:
No in this one.  I just wanted to open this PR for initial comments.

app/assets/stylesheets/amp_styles.scss
```diff
@@ -0,0 +1,166 @@
...
+  margin-bottom: 1rem;
+  color: $color-green-secondary;
+}
+
+h2 {
+  font-size: 1.4rem;
+  line-height: 2rem;
+  margin-top: 1.5rem;
+  margin-bottom: 1rem;
+  color: $color-green-dark;
```
slaywell: These headers can all be the same colour (they are _very_ similar shades on the main site, and that's an oversight that we will correct)

slaywell:
This is looking good. I'm impressed that there's so little CSS needed (I think there may be more when we check other articles, as there are other page components that might need a bit of styling). I won't give a thumb up as it's not ready to merge yet, and I haven't been able to test it out myself.

app/assets/stylesheets/amp_styles.scss
shaley91: Which colour should we be going with?

slaywell:
Still looks good to me 👍

app/controllers/concerns/not_found.rb
```diff
@@ -0,0 +1,9 @@
+module NotFound
+  extend ActiveSupport::Concern
+
+  included do
+    def not_found
```
jongilbraith: The `included` block of a concern is to specify method calls to be executed when the concern gets included.

(Instance) method definitions just go in the body of the module.

Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'link_header'
 gem 'mail'
 gem 'mailjet'
 gem 'mas-assets', git: 'git@github.com:moneyadviceservice/mas-assets'
-gem 'meta-tags', github: 'moneyadviceservice/meta-tags'
+gem 'meta-tags', github: 'kpumuk/meta-tags', branch: 'master'
```
jongilbraith: Good to see us get rid of our fork, but can we not just switch to the published gem then?

shaley91: YES.  But it hasn't been released yet. :(

jongilbraith: Ah right, 👍.

In which case probably good to add a brief note about the features we need that aren't in the latest gem.

shaley91: My plan was to ask the gem owner when the new gem will be released, so we don't have to use master.

jongilbraith: Cool, but still good to add a brief note so anyone who looks at it in the meantime understands why we're tracking master.

jongilbraith: Awesome 👍

app/controllers/concerns/not_found.rb
shaley91: Removed the included block and the `extend ActiveSupport::Concern` line.

slaywell:
I think you have your thumbs @shaley91 - setting to 'DEV APPROVED' it doesn't show up in our feed.

PR #1580, opened 2016-10-28
[DEV APPROVED] Removing obsolete SCSS that generates warnings

slaywell:
The Frontend project gives SCSS warnings every time it is started due to SCSS `@extends` that do not extend any selectors. This will error rather than warn in future versions of SASS.

So this PR removes the unnecessary code, and fixes the instances that pointed to an incorrect `clearfix.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1580)

### DIFF
--- a/app/assets/stylesheets/components/common/_directory.scss
+++ b/app/assets/stylesheets/components/common/_directory.scss
@@ -3,7 +3,7 @@
 // Styleguide Directory
 
 .directory {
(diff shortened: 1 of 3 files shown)


johnplayer1982:
:+1:

davidtrussler:
👍

PR #1589, opened 2016-11-07
Fix cucumber scenario by turning off feature flag

tomas-stefano:
Obs.: This test is relying on the feature flat to be disabled,
but on production is enabled. Since there are no disabled-feature
on cucumber I am disabling on the features.yml and creating
an user story to revisit this and do a proper test when the
feature is disabled or enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1589)

### DIFF
--- a/config/features.yml.sample
+++ b/config/features.yml.sample
@@ -4,4 +4,4 @@ features:
   feedback: false
   optimizely: false
   cms_home_page: true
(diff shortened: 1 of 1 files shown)


jongilbraith:
👍

johnplayer1982:
:+1:

PR #1590, opened 2016-11-07
[DEV APPROVED] 7739 - Bower versioning for Yeast

slaywell:
This PR (and the seven recent ones to engines that use Yeast) allows us to use semantic versioning with Yeast.

I have registered Yeast in the Bower registry by running `bower register yeast https://github.com/moneyadviceservice/yeast` 

This means whenever we update Yeast, we also need to create a Git tag (with the version) then this can be referenced by Bower. Then when updating a project that uses Yeast, simply change the version in the `bower.json.erb` file.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1590)


davidtrussler:
Should we add a description of this process into the Yeast documentation?

davidtrussler:
👍

jongilbraith:
Good stuff 👍

slaywell:
@davidtrussler yes we should! Good thinking.

PR #1591, opened 2016-11-07
use rails sanitize gem not html_escape

Tr4pSt3R:
Cross-Site Scripting opens up our application for manipulation in ways we never intended them to be used. This vulnerability puts user data at risk. It also has the potential to ruin the experience of other users of the MAS website. 

See more about this card on [Target Process](https://moneyadviceservice.tpondemand.com/entity/7578)

## Previously
screen shot 2016-11-08 at 10 18 36

## Currently
screen shot 2016-11-08 at 10 19 13

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1591)

### DIFF
--- a/app/views/styleguide/_element.html.erb
+++ b/app/views/styleguide/_element.html.erb
@@ -1,4 +1,4 @@
 <div class="styleguide-element">
   <span class="styleguide-modifier-name"><%= element.name %></span>
-  <%= source.gsub('$modifier_class', element.class_name).html_safe %>
+  <%= sanitize source.gsub('$modifier_class', element.class_name) %>
 </div>


jongilbraith:
Now that I've looked at this in more depth, I think this is definitely a false positive.

The markup assigned to `source` is the contents of the blocks yielded from the `styleguide_section` helper, such as [here](https://github.com/moneyadviceservice/frontend/blob/master/app/views/styleguide/css_basic.html.erb#L15), so it's definitely markup we've written and know we want to display. (Side note, I think we should therefore be using `raw` instead of `html_safe`).

I'd previously speculated that it was read in from comments in the stylesheets, which @pmcostello was then concerned might be read via an http request and therefore at risk of mitm, but it turns out that is just used for the descriptions (which we don't mark as html safe) and I've also confirmed is [definitely read from the filesystem](https://github.com/kneath/kss/blob/master/lib/kss/parser.rb#L13).

jongilbraith:
Ok, now I think about it a bit more(!), it's the `element.class_name` bit that's the risk, but again, this comes from the Kss gem which again, [pulls the data from the stylesheet comments](https://github.com/kneath/kss/blob/b0791708cb397f5e8995a855af0d1a8b0aa89a2e/lib/kss/modifier.rb#L31), from the filesystem.

Tr4pSt3R:
@jongilbraith Using the `sanitize` method from the gem of the same name resolves the issue. It's simple and effective for what we need it for.

jongilbraith:
@Tr4pSt3R it resolves something that isn't an issue to keep an automated tool happy, yes.

Tr4pSt3R:
Closing this PR after a discussion with @jongilbraith && @tomas-stefano that the issue reported by Hakiri should be treated as a _false positive_

PR #1598, opened 2016-11-15
7725 - persist page feedback state in the client

jongilbraith:
Originally intended to be done with a 30 day cookie, after chatting to @slaywell we decided to go with using localStorage. This is feature detected (and the feedback form will not appear at all if not found) but we are satisfied that support is ubiquitous.

localStorage is a simple key:value store, so we store the data as a serialized json object, with a key of `MAS.onPageFeedback.<article-slug>`.

One caveat: we only retrieve the number of likes and dislikes in the response after doing the actual like/dislike. Therefore, if we later on try to resume the process of asking to share or provide feedback, we don't have the figures to hand once that's complete to show them on the thumbs.

After discussing with Ed (who discussed with Mel), we decided to store the figures we get back after voting along with the other state, and display those.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1598)

### DIFF
--- a/app/assets/javascripts/components/OnPageFeedback.js
+++ b/app/assets/javascripts/components/OnPageFeedback.js
@@ -14,6 +14,8 @@ define(['jquery', 'featureDetect', 'DoughBaseComponent', 'common'], function($, 
     this.ajaxUrl = $el.attr('data-dough-on-page-feedback-post');
     this.likeCount = $el.find('[data-dough-feedback-like-count]');
     this.dislikeCount = $el.find('[data-dough-feedback-dislike-count]');
+    this.likeElement = $el.find('[data-dough-feedback-like-count]');
+    this.dislikeElement = $el.find('[data-dough-feedback-dislike-count]');
     this.pageId = window.location.pathname.match(/^.*\/([a-zA-Z0-9-_]+)$/)[1];
   };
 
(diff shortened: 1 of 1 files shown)


tomas-stefano:
👏 👏 👍 👍

slaywell:
Works great - just one suggestion.

app/assets/javascripts/components/OnPageFeedback.js
```diff
@@ -89,15 +102,56 @@ define(['jquery', 'DoughBaseComponent', 'common'], function($, DoughBaseComponen
...
+      time: Date.now(),
+      likes_count: this.likeElement.text(),
+      dislikes_count: this.dislikeElement.text(),
+      last_state: this.state
+    }
+
+    localStorage['MAS.onPageFeedback.' + this.pageId] = JSON.stringify(data);
+  };
+
+  OnPageFeedback.prototype._applyState = function() {
```
slaywell: After discussion with @jongilbraith, this doesn't quite work, but we are going to have a think about the naming involved in this component. But that's not part of this PR.

jongilbraith:
@slaywell refactored once more based on our final conversation about this yesterday.

slaywell:
👍👍

PR #1599, opened 2016-11-15
[DEV APPROVED] - Add CMS API authentication on page feedbacks endpoints

tomas-stefano:
Since I added the token auth on CMS (you can see here: https://github.com/moneyadviceservice/cms/pull/355) I need to add on the client too (this PR!). 

This still requires work on the puppet scripts to add the token environment variable. More details you can see here: https://github.com/moneyadviceservice/scripts/pull/816

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1599)

### DIFF
--- a/lib/core/repository/cms/page_feedback.rb
+++ b/lib/core/repository/cms/page_feedback.rb
@@ -2,6 +2,8 @@ module Core
   module Repository
     module CMS
       class PageFeedback < ::Core::Repository::CMS::CMS
+        MAS_CMS_API_TOKEN = ENV.fetch('MAS_CMS_API_TOKEN')
+
         def create(params)
           request(:post, params)
         end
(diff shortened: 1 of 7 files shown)


lib/core/repository/cms/page_feedback.rb
```diff
@@ -2,6 +2,8 @@ module Core
   module Repository
     module CMS
       class PageFeedback < ::Core::Repository::CMS::CMS
+        MAS_CMS_API_TOKEN = ENV.fetch('MAS_CMS_API_TOKEN')
```
jriga: What happens if MAS_CMS_API_TOKEN is `nil`

tomas-stefano: The app will not be started. But I think that is the desirable outcome because the app should have the environment variable otherwise could expose a security breach on the app. So if the deployment fails you can rollback to a more secure package instead. **But will not fail because it will guaranteed by the scripts repo when it is deployed).

jriga: 👍

jriga:
LGTM 👍

PR #1622, opened 2016-12-08
[DEV APPROVED] 7840 Car campaign CSS refactor

johnplayer1982:
# Car campaign CSS refactor

Optimising the use of CSS for the car campaign, dropping from 356 lines to 153, a total reduction of 203 lines of CSS.

| [Original page](https://www.moneyadviceservice.org.uk/en/campaigns/revealed-the-true-cost-of-buying-a-car) | [Original page](https://www.moneyadviceservice.org.uk/en/campaigns/how-to-look-ahead-when-buying-a-car) |
|-------|-------|
|image|image|

### Tested in the following

**Browsers**

| Chrome | Firefox | Safari | IE8 | IE9 | IE11 | MSEDGE |
|----------|------|-------|------|----|----|---------|
||||||||

**Devices**

| iPhone 5S | iPhone 6 Plus | Samsung Galaxy S5 | Lumia 930 (Windows Phone) |
|-----------|---------------|---------------------|------------------|
|||||

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1622)

### DIFF
--- a/app/assets/stylesheets/layout/page_specific/_car_campaign.scss
+++ b/app/assets/stylesheets/layout/page_specific/_car_campaign.scss
@@ -36,7 +36,7 @@
 .l-car-campaign__trust-message {
   display: none;
   padding: $baseline-unit*3;
-  background-color: $breadcrumbs-background-color;
+  background-color: $color-grey-pale;
   margin-top: -$baseline-unit*3;
   min-width: 350px;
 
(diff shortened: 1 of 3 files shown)


davidtrussler:
The only (minor) difference I can see between the current live version and the refactored one is that the images attached to sections have been moved away from their current position slightly wide of the margins and the separator image is a bit wider. Not sure if that matter much though.

Otherwise looks great.

image

image

davidtrussler:
Since the image sizes is not an issue: 👍

slaywell:
Great work! 👍

PR #1623, opened 2016-12-09
[WIP] 7772 global nav styling

davidtrussler:
This PR represents the work done on the styling of the global nav so far.

It will need to be reviewed for refining the styles since we are working from a prototype rather than full visual design.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1623)

### DIFF
--- a/app/assets/stylesheets/components/common/_global_nav.scss
+++ b/app/assets/stylesheets/components/common/_global_nav.scss
@@ -1,0 +1,214 @@
+// Global Navigation
+//
+// Styleguide Global Navigation
+
+.global-nav {
+  transition-property: transform;
+  transition-duration: 1s;
+  position: fixed;
+  z-index: 5;
+  width: 100%;
+  transform: translateX(100%);
+
+  @include respond-to($mq-m) {
+    transition: none;
+    position: relative;
+    transform: none;
+    background: $color-yellow-light;
+  }
+
+  &.is-active {
+    transform: translateX(0%);
+  }
+}
+
+.global-nav__arrow {
+  position: absolute;
+  width: (18px / 34) * 25;
+  height: 25px;
+
+  @include respond-to($mq-m) {
+    display: none;
+  }
+}
+
+.global-nav__clumps {
+  transition-property: transform;
+  transition-duration: 1s;
+
+  position: absolute;
+  left: 0;
+  transform: translateX(0%);
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  background: $color-yellow-light;
+  font-size: 1rem;
+
+  @include respond-to($mq-m) {
+    @include column(12);
+
+    transition: none;
+    transform: none;
+    position: static;
+    background: transparent;
+    display: inline-block;
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    justify-content: center;
+  }
+
+  &.is-active {
+    transform: translateX(-100%);
+
+    @include respond-to($mq-m) {
+      transform: none;
+    }
+  }
+}
+
+.global-nav__clump {
+  border-bottom: solid 1px rgba(0, 0, 0, 0.375);
+  box-shadow: inset 0 -1px 2px 0 rgba(0, 0, 0, 0.375);
+  position: static;
+  margin: 0;
+  line-height: 1;
+
+  .global-nav__arrow {
+    right: $default-gutter;
+  }
+
+  @include respond-to($mq-m) {
+    border-bottom: none;
+    box-shadow: none;
+    display: flex;
+    flex-direction: column;
+    flex-grow: 1;
+
+    .global-nav__arrow {
+      display: none;
+    }
+  }
+
+  &.is-active {
+    .global-subnav {
+      transform: translateX(100%);
+
+      @include respond-to($mq-m) {
+        display: block;
+        transform: none;
+      }
+    }
+
+    @include respond-to($mq-m) {
+      background: $color-white;
+    }
+  }
+}
+
+.global-nav__clump:last-child {
+  border-bottom: none;
+}
+
+.global-nav__clump:first-child {
+  border-top: none;
+
+  @include respond-to($mq-m) {
+    .global-nav__clump__heading__text:before {
+      border-left: none;
+    }
+  }
+}
+
+.global-nav__clump__heading {
+  @include column(12);
+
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/components/common/_global_nav.scss
```diff
@@ -0,0 +1,359 @@
...
+  padding: $baseline-unit*2 0;
+
+  @include respond-to($mq-l) {
+    display: flex;
+    width: 100%;
+    margin: 0;
+    line-height: 1;
+    text-align: center;
+
+    &:hover {
```
slaywell: Might be worth adding `:focus` as well, even though we will re-investigate this when we look at the behaviour and accessibility of this

slaywell:
Can we make the pull-out section full-width and add a drop shadow as per the prototype? Then I think those changes should be sufficient to call this 'done' for now.

app/assets/stylesheets/components/common/_global_nav.scss
davidtrussler: Yes, I just added the hover state for a quick-win bit of functionality for demo/review purposes. In fact I think this will need to be done with JS adding the "active" class because of the structure.

davidtrussler:
There is a problem (again associated with the structure) that makes it hard (maybe even impossible) to make the dropdown sections full width, at least using CSS. Or at least I couldn't see one and felt it would be better to leave it as it is for the moment. It would be possible to add this as a refinement with JS although that feels a bit of a hack.

davidtrussler:
The other thing that I was considering adding is some equally quick-win JS to update the classes in the mobile version, otherwise you can only see the different sections by hacking at the code in the browser. Which isn't great for demo purposes.

slaywell:
Good idea to add some quick / removable JS for demoing this.

In terms of the full-width, let's pick this up on Monday. I'm sure it's do-able, but I can't figure it out right now!

johnplayer1982:
Pulled this down locally, looks good :+1:

app/views/shared/svg/_icon_sprite.html.erb
```diff
@@ -75,5 +78,9 @@
     </g>
   </symbol>
 
+  <symbol id="svg-icon--tool" viewBox="0 0 331.6 308.66">
```
slaywell: This already exists in master. If we rebase this branch (and also `global-nav-master` from latest `master` we will get this for free. If we leave it in, it's going to cause a merge conflict, I think.

app/assets/stylesheets/components/common/_global_nav.scss
```diff
@@ -0,0 +1,359 @@
+// Global Navigation
+//
+// Styleguide Global Navigation
+
+$primary-background-colour: #fd5; // yellow;
```
slaywell: Shall we add this as well before merging?

slaywell:
Overall this is working really well. I think there are a few things we should do to tidy up the code that shouldn't have any impact on the appearance:

- Some of the classnames are getting pretty unwieldy, e.g. `global-nav__clump__heading__arrow` - couldn't this just be `global-nav__arrow`? I don't think that BEM names should necessarily mirror DOM structure.
- The SCSS file for _global-nav is huge, could we split out everything that is in the 'content' section into another component?

davidtrussler:
No problem with the overlong classnames. 
There is a little bit of interdependency between the top level styles and the content (really just to activate the categories on the mobile view on line 101) but I don't think that need be a problem.

app/assets/stylesheets/components/common/_global_nav_content.scss
```diff
@@ -0,0 +1,223 @@
+// Global Navigation
+//
+// Styleguide Global Navigation
+
+$max-cols: 6;
+$gutter: $default-gutter / 2;
+
+.global-nav__content {
```
slaywell: If this is the name of the new file (the new 'block') should this be `.global-nav-content`?

app/assets/stylesheets/components/common/_global_nav.scss
```diff
@@ -0,0 +1,206 @@
...
+      bottom: $baseline-unit*2;
+      left: -0.5px;
+      border-left: 1px solid;
+      width: 1px;
+    }
+  }
+}
+
+.global-nav__clump__blog-link__text {
+  padding: $baseline-unit*0.5 0;
```
slaywell: I mentioned this to the Cultivate guys recently, so I'll repeat myself here (sorry) - dividing up the baseline unit makes it pointless using the baseline unit - think it would be simpler to just use an non-variable value here?

app/assets/stylesheets/components/common/_global_nav.scss
```diff
@@ -0,0 +1,206 @@
...
+    }
+  }
+}
+
+.global-nav__clump__blog-link {
+  background: $color-mas-blog-blue;
+  border-radius: 4px;
+  color: $color-white;
+  font-weight: 700;
+  padding: $baseline-unit/2 $baseline-unit*2.5;
```
slaywell: My other comment on the baseline unit also applies here

app/assets/stylesheets/components/common/_global_nav_content.scss
```diff
@@ -0,0 +1,223 @@
...
+  width: 32px;
+  height: 32px;
+  left: -37px;
+  top: 5px;
+  border-radius: 50%;
+  background: white;
+  overflow: hidden;
+}
+
+.global-nav__tool-icon {
```
slaywell: There's still classes that either need renaming (if they are within this file) or moving back to the other file, e.g. `.global-nav__categories` should be in the `global_nav.scss` file, or if it's part of the Global Nav Content block then the class name should be `global_nav_content__categories`.

This is the convention we usually follow, and it's going to make it more confusing later on if we're editing this CSS and we're not sure what file a class is in.

Suggestion: could we come up with a better name for `global_nav_content__*` so that it's clearer that it's a separate BEM component from `global_nav`? Perhaps `nav-content__*`, `nav-inner_*`  or `global-subnav__*`?

davidtrussler: @slaywell Yes, I've found it introduced a few layout issues as well so it needs a bit more work. In terms of the naming convention I think we should keep the "global nav" concept, so happy to go with global-subnav.

slaywell:
Looks good! 👍

PR #1631, opened 2016-12-14
[DEV APPROVED] Update metatags gems

shaley91:
This PR does 2 things.

Firstly it fixes the Gemfile.lock so we don't randomly update gems.  The original change was unintentional.

The second change is to update meta-tags to point to the main meta-tags gem rather than the MAS fork.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1631)


Tr4pSt3R:
👍

tomas-stefano:
I tried to review but I already lost the track from all Gemfile.lock changes, so **if all tests pass** LGTM 👍

jongilbraith:
Since this hasn't been stated already, the important information that people need to know about this PR is that the old fork was created so that we could define a length limit on the description metatag, which is now a feature provided by the gem.

Additionally there was a breaking change because there wasn't a length limit on the title metatag, which there is now. This defaults to 70 and so has been set to a large figure due to the fact that we have titles that are longer than that.

jongilbraith:
👍

PR #1632, opened 2016-12-14
Add support for Google Amp

shaley91:
# Google AMP
## What is it?

AMP is a way to build web pages for static content that render fast. To put it simply, AMP is basically a stripped down HTML page which restricts certain HTML, CSS, and Javascript. It’s designed to be used for articles, not across entire sites.
## AMP Requirements:
- Allows only asynchronous scripts
- All CSS must be inlined
- Styles are limited to 50KB
- External resources like images must state their size in HTML
- All Javascript must be out of critical path
- No user written Javascript is allowed, only AMP supported
- Fonts must be loaded with a link tag or a CSS @font-face rule
# Changes in this PR
- We create an `/amp` sub-route for articles that returns a AMP compliant view. 
- We have created a separate Amp Articles controller 
  - This Subclasses `ActionController::Base` rather than `ApplicationController` because we need to limit everything that gets pulled into the view.
- We created a very basic view that only has AMP compliant HTML as well as various other AMP requirements (i.e. importing the AMP JS lib)
- We created a `amp_styles.scss` file.  This includes very little styles because AMP requires that styling is kept bellow 50Kb
- We added a AmpArticle Decorator that replaces video `iframes` with an `amp-youtube` tag that is AMP compliant
- We link AMP Article views with Articles views and vice versa with canonical link tags.

screenshot 2016-10-27 11 11 27

---

This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1577)


shaley91:
This is dependant on https://github.com/moneyadviceservice/frontend/pull/1631 being merged and released first.

johnplayer1982:
:+1: Looks good

KateWilkinson:
👍

app/controllers/amp_articles_controller.rb
```diff
@@ -0,0 +1,22 @@
+class AmpArticlesController < ActionController::Base
+  include NotFound
```
jriga: This controller is exposed to `Cross-Site Request Forgery` according to hakiri because the method`protect_from_forgery` is not used.

shaley91: Should I copy this line: https://github.com/moneyadviceservice/frontend/blob/master/app/controllers/application_controller.rb#L2

jriga: Yes that should do the trick

jriga:
👍

PR #1635, opened 2016-12-19
7747 - consume the clumps api to populate the global nav

jongilbraith:
Pretty much what it says on the tin.

One thing to mention is that we're not doing anything with the clump link style attribute.

Once we've got more specific specs on that we can add the appropriate logic to customise the view, as well as lock down the permitted values in the CMS too

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1635)


lib/core/repository/clumps/cms.rb
```diff
@@ -0,0 +1,25 @@
+module Core::Repository
+  module Clumps
+    class CMS < Core::Repository::Base
+      def initialize(options = {})
+        self.connection = Core::Registry::Connection[:cms]
+      end
+
+      def all
+        clumps_url = '/%{locale}/clumps.json' % { locale: I18n.locale }
```
jongilbraith: @tomas-stefano yes, we had this discussion the other day.

I was planning on coming back to it as a separate task to tidy these up as the CMS also needs changing (this one doesn't work under both, like some of the api end points).

tomas-stefano: Ok. Sounds good 👍

tomas-stefano:
👍 👍

jriga:
👍

jongilbraith:
@tomas-stefano @jriga thanks both

PR #1637, opened 2016-12-22
[DEV APPROVED] 7849 - refactor strategy for managing VCR cassettes

jongilbraith:
I've wanted to try and find a better approach for this for some time, but the need to do this now was spurred on by the global nav project. It's adding an api request to every page load leads to problems trying to find a nice way to add this to the test suite.

In our request specs we've been making use of explicit `VCR.use_cassette` blocks around specific examples. Clearly this wasn't going to be a practical approach for this problem.

For our cucumber features, we had a reasonably dynamic approach which was able to discern the repository that was making the api call and create a path using that, but the approach I ended up with for the specs had some advantages so I ended up refactoring the features similarly.

In short, this uses VCR's `around_http_request` configuration option to define the cassette path for every single request, which is simply made up of a root directory (cms and google custom search requests get their own directories, then any other gets a directory named according to the request host), within that a directory named according to the http verb, then the rest of the hierarchy is that of the request path.

Cucumber needs a little extra jiggery pokery because of it's threaded nature (I fully confess, I don't 100% grok how that all works but I've tested the behaviour thoroughly and and am convinced the approach I've used does what I expect it to do).

Almost* all the cassettes have now been re-recorded and in theory, you should be able to blow them all away and re-record them without any issue using a CMS with a snapshot of a recent production DB. Adding new tests that make new requests or reuse existing cassettes should be straight forward.

\* Trying to rerecord the cassettes for the features highlighted the fact that we've not changed the tests for retired pages such as action_plans, news, video. Our existing cassettes return data, whereas now they are all redirects to article pages. The original cassettes were preserved for these and a separate ticket will be created to update the features.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1637)


jongilbraith:
Review tip: there are two massive commits in this PR that rerecord all the VCR cassettes, but if you inspect the rest of the commits individually then it shouldn't be too overwhelming.

tomas-stefano:
👍

Tr4pSt3R:
👍

rnalexander:
:+1:  Thanks for the tip on reviewing, that helped!! :shipit:

jongilbraith:
@rnalexander indeed, a hard to see the wood for the trees with all the cassette changes.

My contribution stats are going to go through the roof!

PR #1643, opened 2017-01-03
Recording VCR cassete for clumping request

tomas-stefano:
There it was one cassette that was not on the repository and since there is no CMS running in the build machine the test fails when it tries to do a CMS request. Adding the VCR cassette solves the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1643)


Tr4pSt3R:
👍

davidtrussler:
👍

PR #1647, opened 2017-01-05
[DEV APPROVED]7876 - Adding callout styles

johnplayer1982:
## 7876 - AMP Callout updates

The callout styles on the Google AMP pages needed to be brought inline with the new ones on the MAS site articles.

Google AMP supports some pseudo selectors (:not, :first-of-type, :last-of-type), but not :before or :after.  Because of this it was decided to remove the icons and go with a simpler approach for these mobile pages:

**Without a title**
screen shot 2017-01-05 at 12 36 40

**With a title**
screen shot 2017-01-05 at 12 36 36

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1647)

### DIFF
--- a/app/assets/stylesheets/amp_styles.css.scss
+++ b/app/assets/stylesheets/amp_styles.css.scss
@@ -1,5 +1,6 @@
 $responsive: true;
 @import 'dough/assets/stylesheets/lib/variables';
+@import 'yeast/assets/lib/variables';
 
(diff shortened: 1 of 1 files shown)


slaywell:
👍

rnalexander:
:+1: :shipit:

PR #1649, opened 2017-01-09
[DEV APPROVED] 7883 resolve global nav CI issues

jongilbraith:
Commit 887ed70 builds on the fix @tomas-stefano put in the other day, making it fit in with the rest of the refactored VCR cassettes.

0c393ae I'm hoping will resolve the issues with Go doing the git reset during the build however we'll have to see. My hope is that correcting this, along with a manual reset of the build directory has done the trick.

There is still however a question mark around [this particular build failure](https://go.dev.mas.local/go/tab/build/detail/responsive_commit_uat/152/test_and_build/1/test) which was the actual issue I faced when I was trying to fix this last minute before I left for Xmas. It had apparently resolved itself by the time @tomas-stefano came to look at it, when at which point he encountered a completely different issue (the missing cassette).

Will have to see if that one appears again and address it then if so.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1649)


tomas-stefano:
:+1:

slaywell:
👍

PR #1653, opened 2017-01-12
Fix issue with global nav showing more than subcategories under each tab

jongilbraith:
`category.contents` returns everything in a category - subcategories, articles, guides, etc.

Added a method to the `Category` entity to return just the categories and updated the global nav to use that instead.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1653)


Tr4pSt3R:
👍

johnplayer1982:
:+1:

tomas-stefano:
It is a little late but looks good 👍

PR #1659, opened 2017-01-19
[DEV APPROVED] TP: 7917, Comment: Global Nav: fixes mobile search box functioning

davidtrussler:
Implementing the Global Nav has caused the Search-box dropdown on mobile view to cease to function. This PR re-introduces the code that implements this feature and updates it so that it doesn't break the Global Nav (which was the reason it was originally removed).

## After - Reveals search box

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1659)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -35,6 +35,15 @@ require(['common', 'jquery'], function(MAS, $) {
 
   require(['jquery', 'collapsable'], function($, Collapsable) {
     $(document).ready(function() {
+      // Mobile Search Button
+      new Collapsable({
+        name: 'mobileSearch',
+        closeOffFocus: false,
+        accordion: true,
+        triggerEl: '.mobile-nav__link--search',
+        targetType: 'href',
+      });
+
       // Article Collapsables
       new Collapsable({
         name: 'articleCollapsables',


Tr4pSt3R:
Hi @davidtrussler could you use http://recordit.co to show the new behaviour interaction that this PR will introduce, please?

davidtrussler:
@Tr4pSt3R OK, good point, will do ...

norm:
Maybe add a regression test too? :D

davidtrussler:
@norm I'm not sure what you would expect to see by way of regression test in this instance!

norm:
Well, you broke a thing that worked by adding new stuff. So the regression test would be "test thing works" so you can be relatively confident you'll know if you accidentally break it again in future? :)

I'd approach it by checking out frontend from before the change that broke the thing and write a test to prove the thing works; checkout frontend from after the change, prove the test now breaks (confirming it actually tests what you think it tests); fix the thing that was broken so it works again, prove the test now passes again. Commit fix and test as one.

davidtrussler:
@norm You are right. In this instance the real problem is that there was no test for the original component (which is used fairly widely) otherwise it wouldn't have got broken. My feeling is that should be added but maybe is outside the scope of this PR (which is for a fairly minor bug).

norm:
@davidtrussler I respectfully disagree, fixing a regression should always come with a test where possible.

davidtrussler:
@norm In general terms you are right, but in this instance I don't think there is anything to test. There is a test for the component (to contradict to my earlier comment) but the cause of the bug was that I removed the component from the trigger element. I have essentially now re-introduced this to the same element so I don't feel there is a test that will cover this sequence of events.

norm:
@davidtrussler So there's no need to have a test that reveals that the component was missing and the desired end user functionality broke?

davidtrussler:
@norm I would say that the reason this wasn't caught with a test was because the element in question didn't use what has become our established method of implementing components (e.g. using "data-dough-component" in the markup). My view is that fixing that (which would be worthwhile) is outside the scope of this bug - although it has usefully exposed that problem.

norm:
@davidtrussler Well, I still disagree that adding tests is somehow tangential to fixing bugs.

jongilbraith:
Urgh, sorry @davidtrussler thought I'd picked a time without any open PRs on `global-nav-master` to do the rebase.

Grab me if you need some assistance with rebasing this branch. I've rewritten the history a tiny bit on top of the simple rebase, so it might not be so straight forward.

johnplayer1982:
I'm happy with this, the tests could be more thorough but I agree with David.  
This branch is being merged into the global-nav-master feature branch, there will be a larger review of that and we should consider what type of testing needs to happen then.  For the scope of this task I give it a :+1:

jongilbraith:
👍

PR #1660, opened 2017-01-19
7893 - remove logic to conditionally suppress the mobile menu button in header

jongilbraith:
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1660)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,12 +50,6 @@ class ApplicationController < ActionController::Base
 
   helper_method :display_search_box_in_header?
 
-  def display_menu_button_in_header?
-    true
-  end
-
-  helper_method :display_menu_button_in_header?
-
   def contact_panels_border_top?
     false
   end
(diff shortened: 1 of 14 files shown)


Tr4pSt3R:
@jongilbraith there's a conflict that needs to be solved first. The affected file is `app/views/shared/_mobile_nav.html.erb`. 

Apart from that it's all good. I will thumb it once the conflicts are resolved successfully.

app/controllers/application_controller.rb
```diff
@@ -50,12 +50,6 @@ def display_search_box_in_header?
 
   helper_method :display_search_box_in_header?
 
-  def display_menu_button_in_header?
-    true
-  end
-
-  helper_method :display_menu_button_in_header?
-
```
tomas-stefano: Did you search if the engines uses this helper method? (I am just asking because I faced this problem before on removing a code on the app controller that the engines depends). Perhaps the engines don't use it, but it is good to check.

tomas-stefano:
The code looks good 👍 (besides the checks and the conflicts already pointed)

app/controllers/application_controller.rb
jongilbraith: [Yep](https://github.com/search?q=org%3Amoneyadviceservice+display_menu_button_in_header%3F&type=Code).

And `EmbeddedToolsController` was also one of the controllers that was overriding it for all tools.

jongilbraith:
Branch rebased, addressing *potential* merge conflict. :-)

PR #1665, opened 2017-01-27
[DEV APPROVED] TP: 7894, Comment: Improves accessibility of Global Nav

davidtrussler:
Adds extended keyboard interaction and Aria roles/properties

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1665)

### DIFF
--- a/app/assets/javascripts/components/GlobalNav.js
+++ b/app/assets/javascripts/components/GlobalNav.js
@@ -49,7 +49,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
     this.$globalNav.keydown(function(e) {
       if (e.which === 32 || e.which === 40) {
         e.preventDefault();
-      };
+      }
     });
 
     this.$globalNav.keyup(function(e) {
@@ -72,14 +72,14 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
         case 13:
           if (level === 'top') {
             e.preventDefault();
-            self._openDesktopSubNav(e.target, false);
+            self._openDesktopSubNav(e.target);
             self._moveFocusToSubNav(e.target);
           }
           break;
         // escape
         case 27:
           if (level === 'subnav') {
-            self._closeDesktopSubNav($(e.target).parents('[data-dough-subnav]'), false);
+            self._closeDesktopSubNav($(e.target).parents('[data-dough-subnav]'));
             self._moveTopLevelFocus(e.target);
           }
           break;
@@ -88,7 +88,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
           e.preventDefault();
 
           if (level === 'top') {
-            self._openDesktopSubNav(e.target, false);
+            self._openDesktopSubNav(e.target);
           }
           break;
         // left arrow
@@ -96,7 +96,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
           self._moveTopLevelFocus(e.target, 'left');
 
           if (level === 'subnav') {
-            self._closeDesktopSubNav($(e.target).parents('[data-dough-subnav]'), false);
+            self._closeDesktopSubNav($(e.target).parents('[data-dough-subnav]'));
           }
           break;
         // up arrow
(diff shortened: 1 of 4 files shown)


johnplayer1982:
Really good keyboard interaction! :+1:

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -32,13 +32,177 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
...
 
   /**
+   * Set up keyboard events
+   * Adds enhanced keyboard interation to nav
+   */
+  GlobalNav.prototype._setUpKeyboardEvents = function() {
+    var self = this;
+
+    this.$globalNav.keydown(function(e) {
+      switch (e.which) {
```
slaywell: Do we need this switch statement if both cases are handled identically, and these are the only ones we care about? Would a simple conditional `if (e.which === 40 || e.which === 32)` do the same?

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -32,13 +32,177 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
...
+        // down arrow
+        case 40:
+          e.preventDefault();
+          break;
+        // spacebar
+        case 32:
+          e.preventDefault();
+          break;
+      }
+    })
```
slaywell: Minor, but there's inconsistent semi-coloning throughout, if you run JSHint locally it should point them out.

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -32,13 +32,177 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
...
+  }
+
+  /**
+  * Moves focus of top level nav to left or right
+  */
+  GlobalNav.prototype._moveTopLevelFocus = function(el, dir) {
+    var thisClump = $(el).parents('[data-dough-nav-clump]');
+    var clumps = $(thisClump).siblings('[data-dough-nav-clump]');
+    var firstClump = $(clumps).get()[0];
+    var lastClump = $(clumps).get()[clumps.length - 1];
```
slaywell: I'm not sure that `.get()` is right here. I think this is a [method that's only supported in IE11+ ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get) that's intended to get items from a key/value map. So whilst it works, it's not going to give us our browser support.

Since we're working with jQuery selectors, I think it's better to use  the jQuery methods`.first()` or `.eq()` to filter the set of clumps - https://api.jquery.com/category/traversing/

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -121,7 +285,15 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
...
 
-  GlobalNav.prototype._openDesktopSubNav = function(index) {
+  GlobalNav.prototype._openDesktopSubNav = function(index, delay) {
+    var timeoutDelay;
+
+    if (delay) {
+      timeoutDelay = this.delay;
+    } else {
+      timeoutDelay = 0;
+    }
```
slaywell: I think you can condense these 7 lines to simply `var timeoutDelay = this.delay || 0'`.

If we were using ES2015, this would be a great use case for [default parameters](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Default_parameters).

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -132,21 +304,31 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities'], function($
...
 
-  GlobalNav.prototype._closeDesktopSubNav = function(index) {
+  GlobalNav.prototype._closeDesktopSubNav = function(index, delay) {
+    var timeoutDelay;
+
+    if (delay) {
+      timeoutDelay = this.delay;
+    } else {
+      timeoutDelay = 0;
+    }
```
slaywell: As above, we can make this much less verbose.

app/assets/javascripts/components/GlobalNav.js
davidtrussler: In fact that doesn't quite do the same thing: I've added true and false parameters to determine whether or not we need a delay on the submenu display (i.e. not when it is via a keyboard action). It probably could be less verbose but this breaks some intended functionality.

slaywell:
Keyboard events looking great 👍 👍

PR #1671, opened 2017-02-02
TP7944: Global Nav UI tweaks / styling

slaywell:
- [x] - Styling blog button on mobile menu view
- [x] - Changing animation duration and easing - feels much more snappy
- [x] - Changing colour of text links in menu to black
- [x] - Animating opacity on overlay - @johnplayer1982 this only animates on the way in as when we hide it, we hide it immediately. I think this is OK, but we could work on this more?

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1671)

### DIFF
--- a/app/assets/stylesheets/components/common/_global_nav.scss
+++ b/app/assets/stylesheets/components/common/_global_nav.scss
@@ -2,9 +2,10 @@
 //
 // Styleguide Global Navigation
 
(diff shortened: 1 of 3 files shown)


davidtrussler:
👍  - much improved!

johnplayer1982:
:+1:

PR #1673, opened 2017-02-03
[DEV APPROVED] TP: 7944, Comment: Disables click event on clump headings

davidtrussler:
This disables the click event on the clump headings on desktop view only, which are not links to another area of the site and also have hover events attached to them. It has been added because it is confusing for the user to have both click and hover events on these elements, which cancel each other out.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1673)


jongilbraith:
👍

johnplayer1982:
:+1:

PR #1674, opened 2017-02-03
[DEV APPROVED] 7973 - Remove "directory" block from home page, campaign and error page

jongilbraith:
Edited: removed blurb about tidying of css, instead now everything's been removed.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1674)

PR #1676, opened 2017-02-06
[DEV APPROVED]7979 - Global Nav - responsive homepage restyle

slaywell:
In PR #1674 we are removing the 'Category Directory' from the homepage. This leaves us with a few layout issues with components on the homepage.

This PR:

- Adds a grey background to the 'popular tools and calculators' section that now follows the hero slot
- Changes the way that the homepage handles 'medium' width viewports - see images below - the layout is significantly different. I have also changed some font-sizes and given the hero slot a new layout at medium width viewports.
- Slightly alters the way that the full-width layout appears - the right-hand-side hero image is now constrained within its container

**Screenshots are at around 800px width, but to get a feel for the changes I would recommend local testing / resizing browser window**

New version: 
image

Production: 
image

This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1676)

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_home_top.scss
+++ b/app/assets/stylesheets/components/page_specific/_home_top.scss
@@ -3,7 +3,6 @@
 // Styleguide Home Top
 
 .home-top {
-  padding-bottom: 50%;
   position: relative;
   overflow: hidden;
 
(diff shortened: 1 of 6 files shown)


davidtrussler:
👍 Like it - big improvement

rnalexander:
:+1: :shipit:

PR #1687, opened 2017-02-17
Add custom backtrace silencers to stop custom middlewares obscuring application backtraces

jongilbraith:
This is something that's bugged me for some time on several projects and I only came across an explanation of what's going on the other day. Rails also nowadays includes a template initializer for this sort of thing in newly generated applications.

In short, when we have custom middlewares they produce entries upstream of our application backtrace and unless we tell Rails to ignore them, it won't know that that's not a relevant part to show us in the event of an exception.

**Example: problem with an unexpected nil in my UC engine conversion**

screen shot 2017-02-17 at 15 26 54

screen shot 2017-02-17 at 15 29 35

So if we tell the backtrace silencer to ignore those four custom middlewares:

screen shot 2017-02-17 at 15 26 23

screen shot 2017-02-17 at 15 29 21

Much more helpful!

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1687)

### DIFF
--- a/config/initializers/backtrace_silencers.rb
+++ b/config/initializers/backtrace_silencers.rb
@@ -1,0 +1,5 @@
+# You can add backtrace silencers for libraries that you're using but don't wish to see in your backtraces.
+Rails.backtrace_cleaner.add_silencer { |line| line =~ /^app\/middleware\// }
+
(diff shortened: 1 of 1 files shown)


Tr4pSt3R:
👍 👍

tomas-stefano:
👍 🚢

PR #1688, opened 2017-02-21
[DEV APPROVED] - Add a environment variable for all http requests timeout

tomas-stefano:
The TL; DR we are living on the edge for one request to return 500 every hour on the site or when someone deploys the site.
 
## Context 
 
QA homepage was returning 500 internal server error. The logs shown the site returns 500 but CMS returns 200. Investigating more see it is *exceeding the timeout.*
 
The site makes the request to the CMS and CMS (on QA) takes more than 5 seconds to respond to the categories url (details here: https://cms.qa.dev.mas.local/api/en/categories.json).

Since we have 5 seconds as timeout on the site I decided to "measure" (gross way to measure I would say) how long it is taking *in production*.
It is taking between 4.600 sec to 4.800sec on my machine.
You can run `time curl -XGET https://cms.moneyadviceservice.org.uk/api/en/categories.json` to confirm it what I said. Or use curl with w option that gives the time_total almost look like mentioned above.
 
This PR and https://github.com/moneyadviceservice/scripts/pull/918 addresses the solution.
 

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1688)

### DIFF
--- a/lib/core/connection_factory/http.rb
+++ b/lib/core/connection_factory/http.rb
@@ -5,7 +5,9 @@ require 'faraday/conductivity'
 module Core
   module ConnectionFactory
     class Http
-      def self.build(url, timeout: 5, open_timeout: 5, retries: 2)
+      TIMEOUT = ENV.fetch('FRONTEND_HTTP_REQUEST_TIMEOUT').to_i
+
+      def self.build(url, timeout: TIMEOUT, open_timeout: TIMEOUT, retries: 2)
         options    = { url: url, request: { timeout: timeout, open_timeout: open_timeout } }
         connection = Faraday.new(options) do |faraday|
           faraday.request :json
--- a/spec/lib/core/connection_factory/http_spec.rb
+++ b/spec/lib/core/connection_factory/http_spec.rb
@@ -5,12 +5,12 @@ RSpec.describe Core::ConnectionFactory::Http, '.build' do
     expect(factory).to be_a(Core::Connection::Http)
   end
 
-  it 'has a default timeout of 5 seconds' do
-    expect(factory.options[:timeout]).to eq(5)
+  it 'has a timeout between 5 and 12 seconds' do
+    expect(factory.options[:timeout]).to be_within(5).of(12)
   end
 
-  it 'has a default open timeout of 5 seconds' do
-    expect(factory.options[:open_timeout]).to eq(5)
+  it 'has a open timeout between 5 and 12 seconds' do
+    expect(factory.options[:open_timeout]).to be_within(5).of(12)
   end
 
   it 'defaults to encoding JSON requests' do
--- a/.env
+++ b/.env
@@ -2,6 +2,7 @@ MAS_PUBLIC_WEBSITE_URL=https://www.moneyadviceservice.org.uk/:locale/
 MAS_CMS_URL=http://cms.qa.dev.mas.local
 MAS_CMS_API_TOKEN=mytoken
 MAS_CONTENT_SERVICE_URL=http://localhost:8080/content-service/
+FRONTEND_HTTP_REQUEST_TIMEOUT=10
 GOOGLE_API_KEY=AIzaSyCydfZlpTZlIUUUSbtab5t0oOE-y1GzHy0
 GOOGLE_API_CX_EN=011931388747222259444:razn90vopwc
 GOOGLE_API_CX_CY=011931388747222259444:4thmbtghtbq


pmcostello:
👍

Tr4pSt3R:
👍

spec/lib/core/connection_factory/http_spec.rb
```diff
@@ -6,11 +6,11 @@
   end
 
   it 'has a default timeout of 5 seconds' do
```
norm: Isn't it now a default of 10?

lib/core/connection_factory/http.rb
```diff
@@ -5,7 +5,9 @@
 module Core
   module ConnectionFactory
     class Http
-      def self.build(url, timeout: 5, open_timeout: 5, retries: 2)
+      TIMEOUT = ENV.fetch('FRONTEND_HTTP_REQUEST_TIMEOUT').to_i
```
norm: What value does `TIMEOUT` have if there is no environment variable set? What happens if `FRONTEND_HTTP_REQUEST_TIMEOUT` is set to `banana`?

tomas-stefano: It will be zero.

```
2.4.0 :001 > 'banana'.to_i
 => 0
```

tomas-stefano: If there is no environment variable the application will not start.

norm: And you're happy with both of those outcomes? No need to mitigate them in any way?

tomas-stefano: Yes, I am happy. I don't think anyone will put banana or similar in the http request timeout env var. And yes, I prefer to be strictly to have the env set before the app started than having bugs exploding during app execution.

tomas-stefano: And I happy because if you put 'banana' or similar then the timeout tests will fail. Exactly what I expect too.

norm: That's when running tests. What about when running in production? What does a timeout of 0 actually *do* to the code?

tomas-stefano: It will return an internal server error.

norm: So how about implementing a sensible default value for when the env var is missing or broken? You might be confident about environment variables in production environments always being present and correct, but I am not.

tomas-stefano: First of all I understood your points and **they are valid points**.

But:
1) If the environment variable is missing, the app will not be started, and I am happy with that. And this exactly what I expect from the app itself. There are already some env vars on the site that has the same pattern. If it is missing the app will not start.

2) If someone set to "banana" or something similar *a timeout variable*, the person will be doing on purpose and not in a good faith. The value is already set on the PR on scripts here: https://github.com/moneyadviceservice/scripts/pull/918 And I am prefer to trust that nobody will change to "banana" on the timeout otherwise the person is not even know what they are doing. **Yes, I know, the code is making the assumption about trusting the people that we are working and I trust that no one will put banana or similar**.

And I prefer to don't do the *defensive program around this specific subject*. 
To be fair I don’t technically disagree with defensive program, in fact I agree with some aspects. However, defensive programming in my experience is makes the code less readable and adds a lot of unnecessary code that could be solved in different ways.

Again sorry with the long explanation, I just want to make sure that my opinions was read, heard and understood.

**Just for conclusion, I am really open that** ... **if the whole team agrees to change to be more defensive against "bananas" or similar, I will do it**. @jongilbraith & @Tr4pSt3R What are your thoughts around this?

jongilbraith: I share @tomas-stefano's viewpoint.

Tr4pSt3R:
I concur with @tomas-stefano 's explanation. The variable is already set in the common YAML file on the scripts repo. Whether we should add clauses for catching out mitigating edge-cases I think in this instance we can trust that no one will malicious change the environmental variable into a text. If they do it serves them right if the application does not start.

lib/core/connection_factory/http.rb
jriga: I'm okay with this implementation and not adding defensive code at this level. We should be able to trust that the script project will the correct env var since it goes through the PR thumbing process

jongilbraith:
👍

tomas-stefano:
Thank you all for your patience and time. By the way, @norm your opinions and arguments are always welcome. Saying that, thank you too.

PR #1691, opened 2017-03-01
[DEV APPROVED] 7938 resize promo article images

Tr4pSt3R:
The main work on the CMS side to enable this feature can be seen here: https://github.com/moneyadviceservice/cms/pull/383

[View Card on Target Process](https://moneyadviceservice.tpondemand.com/entity/7938)

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1691)


norm:
This PR as it stands is a 👎 from me. Whilst the end result (as seen in the "Files changed" tab) looks good, the narrative that plays out in each commit is not. There's code and images that get added and then removed again (which is fine during dev, but should be rebased out before issuing a PR), and the individual commits could do with better explanations.

I think it needs to be reworked so that it is one or two good commits. If you're not sure how, I'm happy to pair with you to make it better.

rnalexander:
Please to not merge until have you've had a chance to pair with Norm on how to improve the structure.

Tr4pSt3R:
♻️ 
@norm @rnalexander

lib/core/repository/cms/attribute_builder.rb
```diff
@@ -12,23 +12,21 @@ def self.build(response)
...
 
-        set_description(attributes)
-        set_categories(attributes)
-        set_alternates(attributes)
+        set_description(hsh)
+        set_categories(hsh)
+        set_alternates(hsh)
 
-        attributes
+        hsh
```
norm: What is the purpose of renaming this?

Tr4pSt3R: @norm This is to avoid overloading the method name. I will change it to something more meaning.

spec/lib/core/repository/cms/attribute_builder_spec.rb
```diff
@@ -1,7 +1,87 @@
...
+      let(:hsh) { { 'blocks' => [{ 'identifier' => 'content', 'content' => :bar }] } }
+
+      it 'uses block_composer to set body' do
+        ab.set_body_from_content_block(hsh)
+        expect(hsh['body']).to eq 'bar'
+      end
+    end
+
+    describe '.translate_attributes_from_raw_blocks' do
+      let(:hsh) do
```
tomas-stefano: Can it be renamed to hash, please? IMHO abbreviation it is only useful when it is very very long word.

spec/lib/core/repository/cms/attribute_builder_spec.rb
```diff
@@ -1,7 +1,87 @@
...
+      end
+
+      it 'adds a key without the `raw_` prefix' do
+        ab.translate_attributes_from_raw_blocks(hsh)
+        expect(hsh['tile_1_heading']).to eq hsh['blocks'][0]["content"]
+      end
+    end
+
+    describe '.group_nested_attributes' do
+      let(:hsh) do
```
tomas-stefano: Can it be renamed to hash, please? IMHO abbreviation it is only useful when it is very very long word.

spec/lib/core/repository/cms/attribute_builder_spec.rb
tomas-stefano: Actually, it can be renamed by blocks attributes or even page content itself.

spec/lib/core/repository/cms/attribute_builder_spec.rb
tomas-stefano: Actually, it can be renamed by blocks attributes or even page content itself.

lib/core/repository/cms/attribute_builder.rb
```diff
@@ -12,23 +12,21 @@ def self.build(response)
       end
 
       def attributes
-        attributes = response.body
+        attrs = response.body
```
tomas-stefano: Is there a reason to abbreviate to attrs?

spec/lib/core/repository/cms/attribute_builder_spec.rb
```diff
@@ -1,7 +1,87 @@
+# coding: utf-8
```
tomas-stefano: Is this line really need it?

spec/lib/core/repository/cms/attribute_builder_spec.rb
```diff
@@ -1,7 +1,87 @@
...
 module Core::Repository::CMS
   RSpec.describe AttributeBuilder do
     let(:body) { File.read('spec/fixtures/cms/beginners-guide-to-managing-your-money.json') }
     let(:response) { OpenStruct.new(body: JSON.parse(body)) }
+    let(:ab)  { AttributeBuilder.new(:resource) }
```
tomas-stefano: Can we rename to attribute_builder, please?

spec/lib/core/repository/cms/attribute_builder_spec.rb
```diff
@@ -1,7 +1,87 @@
...
+
+    describe '.set_title_from_label' do
+      let(:hsh) { { 'label' => :foo } }
```
tomas-stefano: Is this hsh variable the response body? If so, can we rename to response_body?

lib/core/repository/cms/attribute_builder.rb
Tr4pSt3R: @tomas-stefano it is abbreviated to _attrs_ as the next best alternative _attributes_ is already used as the method name. Due to how it's heavily used by other methods, it's best to keep the sentiment of _attributes_ even if it's in an abbreviated form. 

```
  30       def set_title_from_label(attributes)
  31         attributes['title'] = attributes['label']
  32       end
  33
  34       def set_body_from_content_block(attributes)
  35         attributes['body'] = BlockComposer.new(attributes['blocks']).to_html
  36       end
  37
  38       def set_description(attributes)
  39         attributes['description'] = attributes['meta_description']
  40       end
  41
  42       def set_categories(attributes)
  43         attributes['categories'] = attributes['category_names']
  44       end
```

app/views/shared/_article_promos.html.erb
```diff
@@ -8,7 +8,12 @@
       <div class="l-promo-articles__item">
         <div class="t-article-promos">
           <%= link_to item['url'], class: 'promo__link' do %>
-            <%= image_tag(item['image'], alt: '', class: 'promo__img') %>
+            <%= image_tag(item['image'],
+              alt: '',
+              sizes: '(max-width: 479px) 100vw, (max-width: 959px) 50vw, 25vw',
```
Tr4pSt3R: @tomas-stefano it's used nowhere else but here. Where is the best place to put this or what is the best practice on handling this, please?

lib/core/repository/cms/attribute_builder.rb
tomas-stefano: Ok, I understand that is heavily used, but it is mutating the same object, so the names could be the same without any problem. I would return to attributes as before because the code becomes better to read.

app/views/shared/_article_promos.html.erb
tomas-stefano: Ok. Is this line very important for the whole feature? What happen with the functionality if I remove this line?

Tr4pSt3R: @tomas-stefano If that line is removed the _srcset_ will not work at the viewports which we expect them to be activated.

lib/core/repository/cms/attribute_builder.rb
Tr4pSt3R: @tomas-stefano I did something like this and @norm had other ideas. I can see why both your ideas are viable. I need a second opinion on this. I am torn between whether to overload the method name or not. 

@jriga wdyt, please?

app/views/shared/_article_promos.html.erb
tomas-stefano: If you remove this line ... will the tests fail to indicate that the feature is not working properly?

lib/core/repository/cms/attribute_builder.rb
norm: @Tr4pSt3R I think you're confusing your notes from our pairing session. My comment on not overloading variable names was based on a piece of code where you reused a variable to contain two different values (that I don't like). This was fine.

rnalexander: I just wanted to say I'm so proud of all of you right now.  😍

Tr4pSt3R: Thanks for clarifying @norm. 
Thanks for pointing that out @tomas-stefano.

app/views/shared/_article_promos.html.erb
Tr4pSt3R: The tests will not fail neither will there be any catastrophic failure. Are you advocating the addition of a test?

tomas-stefano: **I am not advocating anything**. In order to give you the info on the best place to put this line, I need to understand all the implications of it. That's the point of my questions. But I didn't understand one thing: you said that if this line is removed **the srcset will not work** at the viewports and in the other you said there will not be any catastrophic failure. Could not be catastrophic, but is still a failure and the functionality will not work as expected ... right?

pmcostello:
why was the gem rb-readline being added then removed in the last two commits? also could do with a bit more detail in the commit message

Tr4pSt3R:
@pmcostello it's a _fixup_ commit so git automatically adds a commit message to indicate that it's a temporary commit. 

`rb-readline` was added because the system readline library that the Pry debugger uses is mishaving. _Rb-readline_ is a pure Ruby implementation of the _readline_ library. It was temporary to jump into the console to inspect some objects during runtime.

There will no trace of it in the final commit.

Tr4pSt3R:
♻️ @norm @pmcostello @rnalexander @jriga @davidtrussler

jriga:
looks good to me, just needs `git rebase -i --autosquash` to be ready

pmcostello:
ok thanks for the explanation @Tr4pSt3R, squash and merge 👍

Tr4pSt3R:
♻️ @jriga @pmcostello

jriga:
👍

tomas-stefano:
I don't want to sound picky, but please correct if am wrong. Since my understanding of the code that I read:

1) the site makes a request to the CMS and it is *expecting on the json response to have the srcset attribute for every image*.
2) The site renders the *srcset attribute* and add a *sizes attribute* to the img tag.

**Correct if I am wrong but without any of these, the whole feature will not work properly**. RIght?

Conclusion: There is no test that indicates the srcset attribute coming from the cms or even a test to add an expectation the sizes attribute on img tag on the high level spec.

 I saw here that for the [ tag](https://github.com/moneyadviceservice/frontend/pull/1691/commits/068305ccab752d7b6040ae9ec8efe8385e22b3fb#diff-251af4b563462edec4bf87a277a1c6b7R2) you added a helper and the [helper spec](https://github.com/moneyadviceservice/frontend/pull/1691/commits/068305ccab752d7b6040ae9ec8efe8385e22b3fb#diff-3e896d048b2b95ec57e74bdc18eb95cfR1) but I still think that we can test in a high level test (a view spec or even a cucumber spec).

There are some examples on the app for view specs too: https://github.com/moneyadviceservice/frontend/blob/master/spec/views/header_sign_in_box_spec.rb

Or there are examples of cucumber specs that uses VCR to make the request to CMS too on the features dir.

Can we not add a view spec **or** a cucumber spec for this feature?

Tr4pSt3R:
@tomas-stefano have you looked (here)[https://github.com/moneyadviceservice/frontend/pull/1691/files#diff-13f4b45dcf350a578cdf85592ecc7744] and (here)[https://github.com/moneyadviceservice/frontend/pull/1691/files#diff-3e896d048b2b95ec57e74bdc18eb95cf]?

I have put it in [WIP]. Here is a view spec that tests the image _srcset_ size property. The attribute _srcset_ has been tested in this file _spec/lib/core/repository/cms/attribute_builder_spec.rb_. See (here)[https://github.com/moneyadviceservice/frontend/pull/1691/commits/068305ccab752d7b6040ae9ec8efe8385e22b3fb#diff-3e896d048b2b95ec57e74bdc18eb95cfR1]

tomas-stefano:
Since what I saw on the test, the view spec is not testing neither the srcset nor the sizes attribute on img tag. Or there is no integration test that makes the request to the CMS and asserts the the insertion of the srcset attribute (based on the CMS request) and add the sizes attributes as well. Can we pair on this, please?

spec/views/resize_image_spec.rb
```diff
@@ -0,0 +1,17 @@
+RSpec.describe 'shared/_article_promos', type: :view do
+  before do
+		allow(view).to receive(:items_with_image) do
+      [{"heading"=>"Plug those money leaks", "image"=>"https://example.com/MoneyLeaks.jpg", "url"=>"https://example.com/blog", "label"=>"blog", "content"=>"Do you know the most common ways you waste money?" }]
+		end
+
+		allow(view).to receive(:items_without_image) do
+			[{"heading"=>"Estimate the cost of buying a house and moving", "url"=>"https://example.com/blog", "content"=>"There is more to it than just mortgage repayments"}]
+		end
```
tomas-stefano: Can you run rubocop, please? I would fix this identations.

Tr4pSt3R: @tomas-stefano Rubocop is now passing for this file Thanks.

Tr4pSt3R:
♻️ 
@tomas-stefano 

I thought about adding a Cucumber spec but it won't achieve anything so I will leave it. The view spec provides sufficient coverage for this.

tomas-stefano:
Thank you for pairing. But now the view spec is testing the core feature and if you remove the tests will fails, exactly what we expected. Saying that, good work mate! 👍 👍  🚢

PR #1692, opened 2017-03-03
[DEV APPROVED]8057 Related Links nav bug

thecountgs:
There was a fix for IE 8 that caused the sticky nav to flow off the screen to the right. I checked and the fix is not required anymore and removed. I fixed the bug by removing the left positioning call in the js file.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1692)


johnplayer1982:
Checked and can confirm this works, to add some clarity the initial fix was for an IE8 issue and was a minor display bug.  As we have reduced our support for IE8 and the content is still readable we no longer need the fix in place (which was causing a strange issue on modern browsers)

:+1:

davidtrussler:
👍 Looks good!

PR #1709, opened 2017-04-04
[DEV APPROVED] TP: 7996, Comment: Adds alternative navigation for non-js users

davidtrussler:
It is not possible to navigate the current global nav via keyboard commands with JavaScript disabled or if JavaScript does not load. This PR adds an alternative navigation that displays in these circumstances and allows keyboard-only users to be able to navigate the site. 

**Alternate nav on desktop:** 

image

**Alternate nav on mobile:** 

image

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1709)

### DIFF
--- a/app/assets/stylesheets/components/common/_global_nav.scss
+++ b/app/assets/stylesheets/components/common/_global_nav.scss
@@ -134,6 +134,14 @@ $global-nav-transition-duration: 300ms;
   }
 }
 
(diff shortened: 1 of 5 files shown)


thecountgs:
Pulled and tested. 👍

johnplayer1982:
:+1:

PR #1719, opened 2017-04-21
[DEV APPROVED] 7923 global nav add tests

davidtrussler:
This PR adds Karma tests for the Global Nav in the following areas: 

* Test for initialisation of component: check init is calling the appropriate methods. This also slightly refactors GlobalNav.js to put the two one-liners into a separate method. It's is a bit neater I think and helps test the init as a single unit which is probably better practice.
* Tests for binding events. In fact this just tests that the uninitialised class is removed when the component loads
* Test for mobile animation. This should test that the no-transition class is added/removed when the window is resized. I have so far been unable to simulate the resize event in this test so this is left as an empty test (i.e. it passes but only because it isn’t doing anything but can serve as a reminder to work on it at a later date).
* Tests for mobile interaction. This tests: 
    * the nav visibility toggles when the menu button is clicked
    * the dropdown visibility toggles when the clump heading is clicked
    * the dropdown visibility toggles when the subnav heading is clicked
    * the nav is closed when the close button is clicked
    * the nav is closed when the overlay is clicked
* Tests for desktop interaction This tests: 
    * the dropdown is closed on a touch event outside of global nav
    * the dropdown is closed when the user moves the mouse outside of the global nav
    * the dropdown is closed when the user hovers on a clump heading
    * the dropdown is opened when the user touches a clump heading
* Tests for keyboard interactions. This tests: 
    * the enter key opens the dropdown and moves focus into this
    * the spacebar key opens dropdown
    * when at the top level the left arrow key moves focus to the previous element on the top level or wraps if on the first element
    * when at the top level the right arrow key moves focus to the next element on top level or wraps if on the last element
    * when at the top level the down arrow key opens the dropdown
    * the escape key closes dropdown and moves focus back to top level
    * the up arrow key moves focus to the previous element or wraps to the last element if on the first
    * when at the secondary level the down arrow key moves focus to the next element or wraps to the first element if on the last
    * when at the secondary level the left arrow key closes the dropdown and moves focus to the previous top level element or the last element if already on the first
    * when at the secondary level the right arrow key closes the dropdown and moves focus to the next top level element or the first element if already on the last

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1719)

### DIFF
--- a/spec/javascripts/tests/GlobalNav_spec.js
+++ b/spec/javascripts/tests/GlobalNav_spec.js
@@ -1,0 +1,304 @@
+describe('GlobalNav', function() {
+  'use strict';
+
+  beforeEach(function(done) {
+    var self = this;
+
+    requirejs(
+        ['jquery', 'GlobalNav', 'mediaQueries', 'utilities', 'common'],
+        function($, GlobalNav, mediaQueries, utilities, common) {
+          self.$html = $(window.__html__['spec/javascripts/fixtures/GlobalNav.html']).appendTo('body');
+          self.component = self.$html.find('[data-dough-component="GlobalNav"]');
+          self.globalNav = GlobalNav;
+          self.mainContent = self.$html.find('#main');
+          self.mobileNavButton = (self.$html).find('[data-dough-mobile-nav-button]');
+          self.clumps = self.component.find('[data-dough-nav-clumps]');
+          self.subNavHeading = self.component.find('[data-dough-subnav-heading]');
+          self.mobileNavCloseButton = self.component.find('[data-dough-mobile-nav-close]');
+          self.mobileNavOverlay = self.component.find('[data-dough-mobile-nav-overlay]');
+          self.obj = new self.globalNav(self.component, self.component.data('dough-global-nav-config'));
+          self.delay = 250;
+
+          done();
+        }, done);
+  });
+
+  afterEach(function() {
+    this.$html.remove();
+  });
+
+  describe('Initialise component', function() {
+    it('calls the right methods when initialising', function() {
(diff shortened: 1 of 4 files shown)


johnplayer1982:
Very nice!! Tested locally, broke it intentionally to test the fixture, looks good! :+1:

Lets talk testing focus in the next FE team meeting

davidtrussler:
Have added a further commit which sorts out the focus issues on the keyboard events

spec/javascripts/tests/GlobalNav_spec.js
```diff
@@ -0,0 +1,304 @@
...
+    });
+  });
+
+  describe('Mobile animation', function() {
+    beforeEach(function() {
+      this.obj.init();
+    });
+
+    it('toggles the no-transition class when viewport size changes', function() {
+      // This is not functional because I have been unable to simulate the resize event - DT
```
tomas-stefano: Is it possible to add a xit for pending? What is the intended to treat this in the future?

tomas-stefano:
Just a minor comment but the code looks good 👍 Very very nice!!!

spec/javascripts/tests/GlobalNav_spec.js
norm: `window.dispatchEvent(new Event('resize'));` ?

davidtrussler: @tomas-stefano I'm not sure about how to deal with this and ensure it gets done some time. 

@norm The problem with the resize event is that the component uses the media query class that is added to the <html> tag but this doesn't get read in the test scenario.

PR #1705, opened 2017-03-30
[DEV APPROVED] 8056_Endofyear_Warning

thecountgs:
We did similar work last year and I was able to re-use code to deploy this end of year warning for the Redundancy Pay Calculator.

Previous code used for reference:
https://github.com/moneyadviceservice/frontend/commit/9a5f8ad9a037ecb5ce9c4e51a5efb1ef1e2036dc

|Welsh|English|
|------|-------|
|image|image|

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1705)

PR #1722, opened 2017-04-28
[DEV APPROVED] Remove new tax year warning

margOnline:
This pr removes the warning added in pr #1705.

Rather than delete the code which was added back in recently, changing the date ensures the warning is not displayed. There are several options to achieve this result, including setting the whitelist to an empty array.  This pr is to initiate _brief_ discussion on the preferred approach in terms of balancing effort, urgency and future proofing.

We decided to use a yaml file to configure which tools should show budget warnings. If the file is empty, no warning will be displayed. This approach will minimise the developer effort required to enable/disable the message each tax year.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1722)

### DIFF
--- a/app/helpers/budget_warning.rb
+++ b/app/helpers/budget_warning.rb
@@ -1,7 +1,8 @@
 module BudgetWarning
-    ANNOUNCEMENT_DAY   = 06
-    ANNOUNCEMENT_MONTH = 4
-    ANNOUNCEMENT_YEAR  = 2017
+  ANNOUNCEMENT_DAY   = 6
+  ANNOUNCEMENT_MONTH = 4
+  ANNOUNCEMENT_YEAR  = 2018
+  WHITELIST = YAML.load_file('config/budget_warning_tools.yml')
 
   class Announcement
     attr_reader :date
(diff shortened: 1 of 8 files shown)


Tr4pSt3R:
👍 👍

jriga:
👍

app/helpers/budget_warning.rb
```diff
@@ -13,20 +13,18 @@ def initialize
...
   def display_banner_warning?
-    BudgetWarning::Announcement.new.go_live?
+    BudgetWarning::Announcement.new.go_live? && whitelisted?
   end
   module_function :display_banner_warning?
 
   def whitelisted?
-    whitelist = ['/en/tools/redundancy-pay-calculator/',
-                 '/cy/tools/cyfrifiannell-tal-diswyddo/']
+    whitelist = YAML.load_file('config/budget_warning_tools.yml')
```
jriga: we should not trigger some IO, every time it is invoked. We should be able to cache that.

Tr4pSt3R: @margOnline should we move from [DEV APPROVED] to [WIP] until the minor _concern_ above is resolved?

jriga:
👍

PR #1723, opened 2017-05-02
[DEV APPROVED]Remove references to feature toggles

margOnline:
Since we don't use feature toggles, the corresponding gem and config files are redundant.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1723)

PR #1725, opened 2017-05-05
[DEV APPROVED] Fix git feature toggles bug

margOnline:
refer to pr #1723 

As part of the above pr, the support/env.rb file should not require 'feature/testing'

News item tests have been deleted. They were failing because this feature is no longer on the website and redirects to the blog, therefore causing the tests to fail.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1725)


jriga:
👍

johnplayer1982:
:+1:

PR #1737, opened 2017-05-18
[DEV APPROVED] 8120 - Fixing WebChat JS Error

johnplayer1982:
There is currently a JS error on every page apart from the homepage, this is caused by the JS looking for a variable which is undefined as the element it refers to only exists on the homepage (Where the floating chat is restricted to).

| Screenshot of error |
|------------|
|screen shot 2017-05-18 at 15 19 47|

This PR wraps the sections of the JS in the contact panels which refer to this element in the condition:

```
<% if show_floating_chat? %>
```

This way the errors do not appear on any of the other pages, the variable will only be required when the floating chat is active, and will work on every subsequent page we add it to.

**Screenshots of solution**

| Same page as above | Floating chat still appearing on homepage |
|-----|-----|
|screen shot 2017-05-18 at 15 21 00|image|

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1737)

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_floating_chat.scss
+++ b/app/assets/stylesheets/components/page_specific/_floating_chat.scss
@@ -16,6 +16,10 @@
     text-decoration: none;
   }
 
+  &:focus {
+    background-color: $color-white;
+  }
+
   @include respond-to($mq-xs) {
     display: none;
   }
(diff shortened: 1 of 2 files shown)


thecountgs:
👍

margOnline:
👍

PR #1747, opened 2017-06-12
Removing outdated steps on the Readme

tomas-stefano:
We don't need to add the private gem server to /etc/hosts file
anymore and we are not using feature toggles anymore either.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1747)


norm:
👍

pmcostello:
👍

PR #1748, opened 2017-06-13
Add more emphasis on local CMS and include some common gotchas in the README

paulanthonywilson:
Setting up for development documentation:

* Signpost setting up a local CMS a little more
* Adds a section for common gotchas and mitigations encountered while we Cultivars were getting set up on the 12 June 2017.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1748)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,7 @@ The Money Advice Service's responsive website.
 ## Prerequisites
 
 * [Git]
-* [Ruby 2.1.2][Ruby]
-* [Rubygems 2.2.2][Rubygems]
+* [Ruby][Ruby] - see verstion specified in [.ruby-version](.ruby-version)
 * [Node.js][Node]
 * [Bundler]
 * [Bower]
--- a/README.md
+++ b/README.md
@@ -67,9 +67,42 @@ $ foreman s
 
 ### Change CMS URL Path
 
-In development, frontend will use the local CMS for convenience. You can change the MAS_CMS_URL on .env file. Use http://cms.moneyadviceservice.org.uk to point to LIVE content, https://cms.qa.dev.mas.local for testing Or http://localhost:PORT to point to a local running CMS.
+In development, frontend will use the local CMS for convenience. See [CMS repository README](https://github.com/moneyadviceservice/cms/blob/master/README.md) for instructions on setting up a local CMS instance.
+
+You can change the MAS_CMS_URL on .env file. Use http://cms.moneyadviceservice.org.uk to point to LIVE content, https://cms.qa.dev.mas.local for testing Or http://localhost:PORT to point to a local running CMS.
 
 Don't forget to restart the server after the modification.
+
+### Development setup gotchas
+
+#### Contento 404 error
+
+```
+Unable to fetch Footer JSON from Contento error: [#<Core::Connection::Http::ResourceNotFound: the server responded with status 404>] url_prefix: [#<URI::HTTP http://localhost:3000/>]
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,10 +67,43 @@ $ foreman s
 
 ### Change CMS URL Path
 
-In development, frontend will use the local CMS for convenience. You can change the MAS_CMS_URL on .env file. Use http://cms.moneyadviceservice.org.uk to point to LIVE content, https://cms.qa.dev.mas.local for testing Or http://localhost:PORT to point to a local running CMS.
+In development, frontend will use the local CMS for convenience. See [CMS repository README](https://github.com/moneyadviceservice/cms/blob/master/README.md) for instructions on setting up a local CMS instance.
+
+You can change the MAS_CMS_URL on .env file. Use http://cms.moneyadviceservice.org.uk to point to LIVE content, https://cms.qa.dev.mas.local for testing Or http://localhost:PORT to point to a local running CMS.
```
norm: a) *https* for cms.moneyadviceservice.org.uk; and b) whilst I am not speaking authoritatively I would strongly argue that the live service should never be used against a development environment, so I would recommend removing it entirely :)

paulanthonywilson: @norm 👍  for https (I assume)

I can envisage pointing to the production CMS if there was a particularly puzzling issue that was only occurring in production. It's read-only, after all. 

That said, I have no strong opinions either way on documenting that here. Maybe worth checking with @pmcostello,  who added that suggestion, before deleting it?

paulanthonywilson: My mistake - Paul didn't add it, just changed the urls from the previous ones.

shaley91:
Can we remove the ruby versions in Readme as they are not the same as the `.ruby_version` file.

README.md
paulanthonywilson: @norm  Removed the reference to using production CMS in development.

norm: 👍

README.md
```diff
@@ -8,8 +8,7 @@ The Money Advice Service's responsive website.
 ## Prerequisites
 
 * [Git]
-* [Ruby 2.1.2][Ruby]
-* [Rubygems 2.2.2][Rubygems]
+* [Ruby][Ruby] - see verstion specified in [.ruby-version](.ruby-version)
```
Tr4pSt3R: ```see *verstion specified```

* _small_ spelling mistake

README.md
```diff
@@ -67,10 +66,43 @@ $ foreman s
...
+```
+
+The CMS is setup but the database is empty. A solution is to copy the QA CMS database into the CMS, as detailed in the [CMS repository README](https://github.com/moneyadviceservice/cms/blob/master/README.md).
+
+#### Problems loading Dough or Yeast
+
+Assuming you have run `bowndler install`, you may have issues with previous `bower` installations.
+
+```
+rm -rf vendor/assets/bower_components
```
Tr4pSt3R: You can also remove the generated bower json file .i.e ```rm -f bower.json```.

paulanthonywilson: Good call.

Why `-f` though?

README.md
paulanthonywilson: duh!

README.md
Tr4pSt3R: @paulanthonywilson you can do it with/without. I just want it comprehensively vanquished :)

johnplayer1982:
Looks good to me :+1: nice work!

paulanthonywilson:
thanks @Tr4pSt3R @johnplayer1982

PR #1752, opened 2017-06-16
[DEV APPROVED] - 8278 - Show Money Manager in user's "Saved Tools"

igelstorm:
**Target Process ticket**

https://moneyadviceservice.tpondemand.com/restui/board.aspx?acid=13857292912bd09b911bf771e091f219#page=userstory/8278

**Summary**
For a user who has completed the Money Manager questionnaire, show Money Manager in the Saved Tools section of My Account.

**Examples (optional)**
screen shot 2017-06-16 at 12 12 53

**Checklist**

Make sure that when your PR is good to be reviewed you check the below checks:

+ [x] Tests passed.
+ [x] Rubocop is passing for this PR (or any other lint like JShint).
+ [x] Code Climate passed for this PR.
+ [x] Commit history reviewed (clear commit messages).

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1752)

PR #1760, opened 2017-06-29
[DEV APPROVED] - 8303 - Money Manager: Enhance sign-out behaviour

picandocodigo:
[TP Link](https://moneyadviceservice.tpondemand.com/entity/8303)

**Summary**
Currently, when a logged in user signs out using the link in the header, they remain on the page they were on, and any personal information is still visible. The preferred functionality is to be returned to the Money Manager landing page in a signed out state.

This was because the session data wasn't being deleted when signing out.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1760)


features/sign_in.feature
```diff
@@ -32,3 +32,11 @@ Feature: Sign in
     And   I should be at the page I was on
     And   I should receive a "Signed out successfully." notification
 
+  Scenario: Sign out after using Money Manager
```
Tr4pSt3R: @picandocodigo tbh this scenario can be moved to another file as it is a new feature. 

```
#file: features/enhance_signout_behaviour.feature
.
.
.
```

picandocodigo:
@Tr4pSt3R We've updated the PR moving the enhanced sign out to a new feature and gave it a more descriptive name.

Tr4pSt3R:
@picandocodigo I can't find where the changes have been made was it made as part of this commit(`c4b4928`), please?

picandocodigo:
@Tr4pSt3R yes, we amended that commit and added the changes there.

features/step_definitions/money_manager_steps.rb
```diff
@@ -0,0 +1,22 @@
+Given(/^I am on the Money Manager tool$/) do
```
Tr4pSt3R: @picandocodigo it looks great! The name for this feature could be changed to reflect the primary feature which is _Enhance sign-out behaviour_.

picandocodigo: @Tr4pSt3R Since the steps definition is very specific to the Money Manager tool, we named it `money_manager_steps` and named the feature `clear_data_on_signout` because we felt it better explained what is going on. If we add more sign out enhancements we can rename it to `enhance_sign_out_behaviour`.

Tr4pSt3R: Money Manager is the name of the tool rather than a clear description of the purpose for this feature. It's always helpful to match the feature name to the step definitions. 

I.e. `features/clear_data_on_signout.feature` should be expected to have it's step definitions defined within `features/step_definitions/clear_data_on_signout_steps.rb`

features/step_definitions/money_manager_steps.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+Then(/^I should not see any hint of my details when I re-visit the tool$/) do
+  expect(find_field(id: 'questionnaire_received_first_payment_true')).not_to be_checked
+  expect(find_field(id: 'questionnaire_received_first_payment_false')).not_to be_checked
+end
+
+Then(/^I should be at the Money Manager landing page$/) do
+  expect(page.current_path).to eql('/en/tools/money-manager/')
+end
```
tomas-stefano: These "find_fields" expressions, should they belong to the page object?

features/step_definitions/money_manager_steps.rb
picandocodigo: Right! I see the point there. We've renamed it accordingly.

features/step_definitions/money_manager_steps.rb
picandocodigo: Maybe, are you referring to something like `money_manager_page`? We looked at how things were being done on other tests and were wondering about where these helpers were being created but we assume they are being auto generated by the DSL. Should we then use something like `money_manager_page.find_field`?

features/step_definitions/clear_data_on_signout_steps.rb
```diff
@@ -0,0 +1,22 @@
+Given(/^I am on the Money Manager tool$/) do
+  visit('/en/tools/money-manager/')
+end
+
+Then(/^I complete the series of questions$/) do
+  find_field(id: 'questionnaire_received_first_payment_true').click
```
Tr4pSt3R: Hey @picandocodigo how do you feel about using a [_page object model DSL for Capybara_](https://github.com/natritmeyer/site_prism) for cleaning up some of this code, please?

igelstorm: We're refactoring this now, using the other specs as a guide. It looks like the other specs are using some kind of helper (e.g. `debt_management_page`) instead of instantiating the SitePrism page directly (e.g. `DebtManagement.new`, which is what the SitePrism readme suggests), but we can't see where these are coming from. We assumed they're generated automatically by some metaprogramming somewhere, but we don't have a `money_manager_page` available after creating a MoneyManager class. Could you point us towards any code or docs that will help us understand this?

tomas-stefano: @igelstorm that's a good spot! Even if I disagree with the existing code of the frontend ... the "magic" place to look is on https://github.com/moneyadviceservice/frontend/blob/master/features/support/world/pages.rb

picandocodigo:
Thanks @Tr4pSt3R and @tomas-stefano, we've refactored the tests using site_prism and updated the commit.

features/step_definitions/clear_data_on_signout_steps.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+Then(/^I complete the series of questions$/) do
+  money_manager_page.received_first_payment_true.click
+  money_manager_page.single_or_in_couple_single.click
+  money_manager_page.submit.click
+end
+
+Given(/^I get my results$/) do
+  expect(page.current_path).to eql('/en/tools/money-manager/to-read')
```
Tr4pSt3R: The expect page here should be `money_manager_page`, please.

features/step_definitions/clear_data_on_signout_steps.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+Then(/^I should not see any hint of my details when I re-visit the tool$/) do
+  expect(money_manager_page.received_first_payment_true).not_to be_checked
+  expect(money_manager_page.received_first_payment_false).not_to be_checked
+end
+
+Then(/^I should be at the Money Manager landing page$/) do
+  expect(page.current_path).to eql(money_manager_page.url)
```
Tr4pSt3R: The same thing here, the expected page should be `managing_money_page`. Looking excellent though.

tomas-stefano:
👏  👍 👍

features/support/ui/pages/money_manager.rb
```diff
@@ -0,0 +1,16 @@
+require_relative '../page'
+
+module UI::Pages
+  class MoneyManager < UI::Page
+    set_url '/en/tools/money-manager/'
+
+    element :received_first_payment_true,
+            '#questionnaire_received_first_payment_true'
```
Tr4pSt3R: In order to de-couple test from actual CSS implementations we tend to add descriptive classes to the field and then reference it in the test. 

Within the view file we add this class `.t-questionnaire_received_first_payment_true` to the field. And then set it as the reference in the page objects as follows, `element :received_first_payment_true, '.t-questionnaire_received_first_payment_true'

features/support/ui/pages/money_manager.rb
```diff
@@ -0,0 +1,16 @@
...
+    element :received_first_payment_false,
+            '#questionnaire_received_first_payment_false'
```
Tr4pSt3R: The same thing applies here when it comes to de-coupling actual CSS implementations.

features/support/ui/pages/money_manager.rb
```diff
@@ -0,0 +1,16 @@
...
+    element :single_or_in_couple_single,
+            '#questionnaire_single_or_in_couple_single'
```
Tr4pSt3R: Same thing here with this id `#questionnaire_single_or_in_couple_single`.

Tr4pSt3R:
CodeClimate is also failing. @picandocodigo or @igelstorm has the Code Climate credentials been given to you please?

igelstorm:
@Tr4pSt3R Yeah, we have access, thanks. We had exactly the same CodeClimate issues appear in another PR the other week, and @tomas-stefano said not to worry about it: https://github.com/moneyadviceservice/frontend/pull/1752#issuecomment-309439553

At that time it seemed totally inexplicable because we hadn't even touched those files. Now we have, but I still don't really see how you could avoid that duplication in a nice way without a large refactor. (And I guess Tomas's point about Code Climate being intended more for code than specs still applies.)

features/step_definitions/clear_data_on_signout_steps.rb
picandocodigo: In this story we specifically need to check that we see the 'to-read' page which is different from `money_manager_page`'s url. Is there a way we can do that with the same object? Or should we create a `managing_money_page` object for that?

features/step_definitions/clear_data_on_signout_steps.rb
picandocodigo: I'm confused about this comment. Do you mean the `managing_money_page` as in the `to-read` url or something else? In this case we are testing that we're at the root of Money Manager, so the url is just `/en/tools/money-manager/`. Should we change the parameter for `expect` or the parameter for `eql`?

features/support/ui/pages/money_manager.rb
igelstorm: To be honest, I'm not sure what benefits that would give compared to a unique id, in terms of decoupling. I can see the value if you'd otherwise have to select by CSS classes that aren't unique, or that are cosmetic and could change, but when there is already an ID and it's unique, that seems like the best option.

features/step_definitions/clear_data_on_signout_steps.rb
Tr4pSt3R: @picandocodigo ah! pardon me. This one is fine 👍

features/step_definitions/clear_data_on_signout_steps.rb
Tr4pSt3R: @picandocodigo this is fine too 👍

features/support/ui/pages/money_manager.rb
Tr4pSt3R: @margOnline @willhall88 @tomas-stefano what do you think about changing the standards on the naming convention for identifying models for the page object elements. 

The idea is to add `.t-` to help identify classes that are used for test purposes rather than styling. 

My question is, is it okay to drop this _standard_ given that Rails automatically creates unique ids for some elements?

Please give your suggestions so that we can get this out of the door.

features/support/ui/pages/money_manager.rb
Tr4pSt3R: @igelstorm We are currently looking into whether to drop or uphold these standards. I do agree with you though

features/support/ui/pages/money_manager.rb
tomas-stefano: Is this an input? Why not using input[name='foo'] (foo it is an example).

pmcostello:
👍

Tr4pSt3R:
👍

PR #1765, opened 2017-07-07
[DEV APPROVED] 8324 Home Buying Checklist

thecountgs:
This is a PR for ticket [#8324](https://moneyadviceservice.tpondemand.com/entity/8324) Home Buying Checklist.

# Goal

- Build a collection of inputs that tally up the numerical total without reloading the page. 
- Display unique informational messages via focus and blur events for each input.
- Display unique informational messages via collapsable component.
- Make the UX/design clean and easy to use.
- Allow for printing and storing via cookie.
- Display simplified mobile view

Tasks completed have been checked. 

- [x] Create page and add Route
- [x] Add markup
- [x] Add collapsable components
- [x] Add form inputs with field validation
- [x] Add javascript calculation
- [x] Create responsive mobile and desktop views 
- [x] Add functionality to store form data in LocalStorage
- [x] Create print stylesheet

| Desktop | Mobile |
|---------|--------|
|image | image |

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1765)

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
+++ b/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
@@ -161,7 +161,7 @@ input.home-buying-checklist__form-input--green {
 
   @include respond-to($mq-s) {
     @include column(2);
-    dislay: block;
+    display: block;
   }
 }
 
@@ -170,7 +170,7 @@ input.home-buying-checklist__form-input--green {
 
   @include respond-to($mq-s) {
     @include column(6);
-    dislay: block;
+    display: block;
   }
 }
 
@@ -180,7 +180,7 @@ input.home-buying-checklist__form-input--green {
 
   @include respond-to($mq-s) {
     @include column(4);
-    dislay: block;
+    display: block;
   }
 }
 
@@ -212,7 +212,7 @@ input.home-buying-checklist__form-input--green {
 
   @include respond-to($mq-s) {
     @include column(8);
-    dislay: block;
+    display: block;
   }
 }
 
@@ -222,6 +222,6 @@ input.home-buying-checklist__form-input--green {
 
   @include respond-to($mq-s) {
     @include column(4);
-    dislay: block;
+    display: block;
   }  
 }
--- a/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
+++ b/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
@@ -60,28 +60,6 @@ input.home-buying-checklist__form-input--green {
   }
 }
 
-.home-buying-checklist__collapsable {
-  @include column(12);
-}
-
-.home-buying-checklist__collapsable-trigger {
-  display: inline-block;
-  line-height: 1em;
-  font-style: italic;
-  text-align: left;
-  display: inline-block;
-  position: relative;
-  margin: 0;
-  font-size: 1rem;
-  font-weight:500;
-}
-
-.home-buying-checklist__collapsable-target {
-  display: inline-block;
-  background: $color-green-pale;
-  padding: $baseline-unit * 3;
-}
-
 .home-buying-checklist__form {
   @include column(12);
   @extend %clearfix;
@@ -93,6 +71,7 @@ input.home-buying-checklist__form-input--green {
 .home-buying-checklist__form-description {
   @include column(12);
   @extend %clearfix;
+  position: relative;
 }
 
 .home-buying-checklist__callout-active {
--- a/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
+++ b/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
@@ -223,5 +223,32 @@ input.home-buying-checklist__form-input--green {
   @include respond-to($mq-s) {
     @include column(4);
     display: block;
-  }  
+  }
+}
+
+@media print{
+  input.home-buying-checklist__form-input,
+  input.home-buying-checklist__form-input--green {
+    &::-webkit-outer-spin-button,
+    &::-webkit-inner-spin-button {
+      -webkit-appearance: none;
+      margin: 0;
+    }
+
+    -moz-appearance:textfield;
+    border: none;
+    padding: 0;
+    height: auto;
+  }
+
+  .home-buying-checklist__collapsable {
+    display: none;
+  }
+
+  .home-buying-checklist__form-left,
+  .home-buying-checklist__form-right,
+  .home-buying-checklist__form-results-left,
+  .home-buying-checklist__form-results-right {
+    @include column(6);
+  }
 }
--- a/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
+++ b/app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
@@ -1,4 +1,4 @@
-form {
+.home-buying-checklist__form {
   @include column(12);
   @extend %clearfix;
 }
@@ -66,15 +66,6 @@ input.home-buying-checklist__form-input--green {
 
 .home-buying-checklist__collapsable-trigger {
   display: inline-block;
-}
-
-.home-buying-checklist__collapsable-target {
-  display: inline-block;
-  background: $color-green-pale;
-  padding: $baseline-unit * 3;
-}
-
-.icon-info {
   line-height: 1em;
   font-style: italic;
   text-align: left;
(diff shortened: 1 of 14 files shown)


thecountgs:
There is an issue with the indentation on the html view in my editor was ok but when uploaded to github the indentation became janky in parts. Apologies, just spotted this.

app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
```diff
@@ -0,0 +1,254 @@
+form {
```
igelstorm: This should be something like `.home-buying-checklist__form`.

igelstorm: (This is just a note to ourselves to fix it - nobody at MAS needs to care. Just in case anyone gets an email notification...)

margOnline:
the spacing/tabbing for the view file, app/views/home_buying_checklist/show.html.erb is inconsistent, please can this be addressed

margOnline:
there are no tests 😱

app/controllers/home_buying_checklist_controller.rb
```diff
@@ -0,0 +1,11 @@
+class HomeBuyingChecklistController < ApplicationController
+
+  def display_skip_to_main_navigation?
+    false
+  end
+
+  def show
+
```
Tr4pSt3R: Put this on online perhaps? `def show; end`

tomas-stefano:
We have controller code with no tests and the javascript does not have tests either. Can we add those, please?

igelstorm:
I've taken the WIP off, as we have addressed the issues raised here and on [our corresponding Trello card](https://trello.com/c/i6XbCAys).

@tomas-stefano You mentioned controller tests - what kind of tests would you expect to see for something like this? Just a controller spec that asserts the right view is rendered, or something more? We have added tests for the JS that's on the page, but since the controller has no logic in it, I wasn't sure what kind of tests would provide any value.

app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
```diff
@@ -0,0 +1,215 @@
...
+
+
+.home-buying-checklist__form-description {
+  @include column(12);
+  @extend %clearfix;
+  position: relative;
+}
+
+.home-buying-checklist__callout-active {
+    display: block;
```
johnplayer1982: Please change indentation to 2 space

app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
```diff
@@ -0,0 +1,215 @@
...
+}
+.home-buying-checklist__callout-inactive {
+    display: none;
```
johnplayer1982: Same here, indentation

app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
```diff
@@ -0,0 +1,215 @@
...
+  @extend %clearfix;
+  font-weight: 500;
+  font-size: 1.125rem;
+  border-bottom: 1px solid $color-bluegrey-light;
+  background: $color-green-pale;
+}
+
+.home-buying-checklist__form-row {
+	@extend %clearfix;
+	border-bottom: 1px solid $color-bluegrey-light;
```
johnplayer1982: Indentation also here please

app/assets/stylesheets/components/page_specific/_home_buying_checklist.scss
```diff
@@ -0,0 +1,215 @@
...
+
+.home-buying-checklist__form-header-left,
+.home-buying-checklist__form-header-center,
+.home-buying-checklist__form-header-right,
+.home-buying-checklist__form-left,
+.home-buying-checklist__form-center,
+.home-buying-checklist__form-right,
+.home-buying-checklist__form-results-left,
+.home-buying-checklist__form-results-right {
+	padding: $baseline-unit * 3;
```
johnplayer1982: Last indentation one

app/views/home_buying_checklist/show.html.erb
```diff
@@ -0,0 +1,371 @@
...
+
+      <div class="home-buying-checklist__form-row--gray">
+        <div class="home-buying-checklist__form-results-left">
+          <div class="home-buying-checklist__form-label">
+            Your total upfront cost:
+          </div>
+        </div>
+
+        <div class="home-buying-checklist__form-results-right" aria-live="polite">
+          £ <span data-dough-input-result>0</span>
```
johnplayer1982: This may not work as intended, could you move the ``aria-live`` attribute to the span within and add some additional markup, this is to ensure that a screen reader will read as "Total £value" rather than just the number

```
<span aria-hidden="true">£</span>
<span data-dough-input-result aria-live="polite">
  <span class="visually-hidden">Total £</span>
  0
</span>
```

tomas-stefano:
@igelstorm if you remove the method 'display_skip_to_main_navigation?' what happens? Will this be a bug? If so, is not better to add one test that garantees that this is false?

picandocodigo:
@tomas-stefano `display_skip_to_main_navigation` is being used to display a link to a `#primary-nav` navigation which is present only in some pages. So some controllers display it, others don't. It's an accessibility link for screen readers apparently, since it's invisible. We didn't find any tests for it on the other controllers that override this method.

app/views/home_buying_checklist/show.html.erb
igelstorm: We've tested this with a screen reader and made some changes.

It seems to be fine to use `aria-live` on a div - it doesn't need to be a span. You're totally right about ensuring it doesn't just read the number - we moved the `aria-live` to the parent element, so now it reads "Your total upfront cost: X pounds" whenever it changes. Setting `aria-atomic` to true ensures that it reads the entire contents of the div instead of just the element that updates.

Every day is a school day - it's been interesting to learn more about screen reader accessibility.

tomas-stefano:
@picandocodigo after talking with @davidtrussler and investigating more about what it is doing and what happens if this is true, I think it is better that we add a test to guarantee it is false. 

The reason is: If I change to true, there will be an accessibility link to go to the main navigation which on this tool/page/PR there is no main navigation, which can confuse users that uses the accessibility links to click on the link without going nowhere. Other thing is to be careful with the though "the existing code is not doing this". There are some parts on the site, that the existent code should not be something to be based on, only consider it. 

Sorry to be stubborn on this issue, the PR looks great, but can we add a test for this matter, please?

picandocodigo:
@tomas-stefano that's fine, we've just updated the PR with a test for that controller method.

johnplayer1982:
This is great, gets a front end :+1:

spec/javascripts/tests/HomeBuyingChecklist_spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+    it('persists values when they are changed', function() {
+      var firstInput = this.component.find('[data-dough-input-value]:first');
+      firstInput.val(3);
+      firstInput.trigger("change");
+
+      expect(localStorage.getItem(this.test_key)).to.equal('3');
+    });
+  });
+});
```
margOnline: 👍  for tests!

app/views/home_buying_checklist/show.html.erb
```diff
@@ -0,0 +1,371 @@
...
+          </div>
+        </div>
+
+        <div class="home-buying-checklist__form-results-right">
+          £ <span data-dough-input-result>0</span>
+        </div>
+      </div>
+    </div>
+  </form>
+</div>
```
margOnline: This is quite a long file which makes me wonder if there is an opportunity to refactor in some way:

using partials - perhaps for the popup_tips or the checklist__form-row or

is it possible to provide an array of the different checklist items to the view. The view could then loop through the array, rather than each one being individually hardcoded, vis a vis to how you've streamlined checklist.js?

gotbadger: @margOnline Ideally we would do something like this but the view isn't translated at all which means unless we use a translation we would have to use an array of hashes or some such and I don't think that would be much cleaner.

For me I think the point at which this would become a full featured tool with translations would be the point at which it would make sense to refactor this. Alternatively if you think it is important right now we can refactor it and add translations leaving the welsh versions as english for now.

margOnline: @gotbadger, makes sense.

margOnline:
👍

tomas-stefano:
👍

gotbadger:
Some more work has been requested on this branch so removed `[DEV APPROVED]`

PR #1789, opened 2017-08-18
[DEV APPROVED] 8391 - Broken page should show 404 error

igelstorm:
[TP link](https://moneyadviceservice.tpondemand.com/entity/8391)

**Summary**
Visiting https://www.moneyadviceservice.org.uk/en/corporate_categories/blah (where 'blah' can be by any nonexistent category) returns a 500 error (`Core::Repository::Base::RequestError 'Unable to fetch Category JSON from Contento'` is raised). A 404 should be returned instead.

This is ultimately because `Core::Repository::Categories::CMS` errors when a category doesn't exist in the CMS, and this error is never caught. Changing this behaviour deeper in the system could potentially impact on other code that depends on it, so we just rescue it in the controller and show a 404.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1789)

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -15,6 +15,8 @@ class CategoriesController < ApplicationController
     @breadcrumbs = BreadcrumbTrail.build(@category, category_tree)
 
     assign_active_categories(@category)
+  rescue Core::Repository::Base::RequestError
+    not_found
   end
 
   private
--- a/features/cassettes/CMS/get/api/en/categories/nonexistent-category_json.yml
+++ b/features/cassettes/CMS/get/api/en/categories/nonexistent-category_json.yml
@@ -1,0 +1,51 @@
+---
+http_interactions:
+- request:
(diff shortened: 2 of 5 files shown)


igelstorm:
Good point, @Tr4pSt3R - I've updated the spec to reflect what actually happens when a category is missing.

tomas-stefano:
Can you add a cucumber step for this bug? Adding a cucumber test will be easier to implement [this TP user story](https://moneyadviceservice.tpondemand.com/entity/8097) whenever we done it.

igelstorm:
@tomas-stefano Does this look ok? https://github.com/moneyadviceservice/frontend/pull/1789/commits/89de99e6fab9a3e5d4591df8d3ebbf7c771dc195

tomas-stefano:
@igelstorm looks great!! 👍 👍

Tr4pSt3R:
👍

PR #1790, opened 2017-08-21
[DEV APPROVED] 8364 - Fix Welsh links

igelstorm:
[TP link](https://moneyadviceservice.tpondemand.com/entity/8364)

**Summary**
This fixes a number of incorrect/missing Welsh links on https://www.moneyadviceservice.org.uk/en/corporate/.

The ticket pointed out one of these - the "Media centre" URL that was used works in English, but has no corresponding Welsh page and thus caused an error when the "Welsh" link is clicked.

In the process of fixing this, I found that the view had hardwired English URLs for several links, even though Welsh versions exist. So this PR also addresses that by using path helpers, and moving the link slugs into translation files.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1790)

### DIFF
--- a/app/views/corporate/index.html.erb
+++ b/app/views/corporate/index.html.erb
@@ -8,46 +8,46 @@
 
       <div class="l-promo-corporate">
         <div class="l-promo-corporate__item">
-          <a class="promo__link" href="/en/corporate_categories/about-us">
+          <%= link_to corporate_article_path(t('corporate_home.about_us.link')), class: 'promo__link' do %>
             <%= image_tag('corporate/about_us_promo.jpg', class:'promo__img') %>
             <h2 aria-level="2" class="promo__heading" role="heading"><%= t('corporate_home.about_us.heading') %></h2>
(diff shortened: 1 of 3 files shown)


app/views/corporate/index.html.erb
```diff
@@ -15,7 +15,7 @@
           </a>
         </div>
         <div class="l-promo-corporate__item">
-          <a class="promo__link" href="/en/corporate_categories/media-centre">
+          <a class="promo__link" href="/en/corporate/media-centre">
```
tomas-stefano: I understand the static url change, but is it possible to use rails routing? Does `corporate_article_path('media-centre')` works?

igelstorm: @tomas-stefano I've done this, and also addressed a problem I missed the first time around (missing Welsh links!). I've updated the PR description since this changes the scope quite a lot.

tomas-stefano:
LGTM 👍

margOnline:
👍

PR #1791, opened 2017-08-21
[DEV APPROVED] - 8457 - Remove static page with broken links

igelstorm:
[TP link](https://moneyadviceservice.tpondemand.com/entity/8457)

**Summary**
The above bug in TP reported that an [old, nonexistent version of Budget Planner](https://www.moneyadviceservice.org.uk/en/tools/budget-planner/short-form/edit/money-in) is still linked to from the site.

The broken link was found in this static page, and it seems like the appropriate fix is to remove the page altogether, since the information in it seems to be entirely outdated. I'd appreciate any comments about whether removing this page is likely to have any wider implications.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1791)


thecountgs:
👍

margOnline:
@tomas-stefano do you know if removing this page has wider implications?

tomas-stefano:
As far I am aware of there is no impact. However I recommend investigate places that links to this removed page. Like: https://github.com/moneyadviceservice/budget_planner/blob/49f138cb295d5913b81db74cb3eb9ab61aa57b85/app/views/budget_planner/deprecation_notifier/bp3.html.erb#L225

igelstorm:
@tomas-stefano I found that file as well - it seems to be an email that was sent out manually using a [rake task](https://github.com/moneyadviceservice/budget_planner/blob/master/lib/tasks/deprecation.rake) at some point. We could remove it, but then we should also remove the mailer and the rake task. I wasn't sure whether this was worth doing since it won't be seen unless someone manually runs the task again. What do you think?

tomas-stefano:
The task will not be runned again so IMHO the mailer, rake task and the email view could be deleted as well since the short form does not exist anymore.

igelstorm:
@tomas-stefano OK, I've made a PR for removing these files: https://github.com/moneyadviceservice/budget_planner/pull/483

tomas-stefano:
Good Now 👍 👍

PR #1792, opened 2017-08-21
8324 add target to links so they open in same window

gotbadger:
[TP Link](https://moneyadviceservice.tpondemand.com/entity/8324).

This PR adds some changes to how links are handled on home buying checklist PR(https://github.com/moneyadviceservice/frontend/pull/1765). 
We created this as a separate PR so changes are easier to review.

Links now open in a new window but with the same target so that a new tab is only opened once

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1792)


picandocodigo:
👍

PR #1795, opened 2017-08-23
[DEV APPROVED] 8433 - Replacing old wpcc with the new one

tomas-stefano:
[TP Ticket](https://moneyadviceservice.tpondemand.com/entity/8433)

**Summary**

In order to release the new WPCC to prod, we need to take the old WPCC off and release the new one. 

**Solution**

In order to replace the following was necessary:

* The contribution_calculator (the old wpcc) gem should be removed.
* Wpcc using git until we have the final version bumped.
* Remove authentication from the wpcc in order to be open to the users.
* Remove the robots reference because there is no /en/tools/wpcc link anymore.

There is a pending part on this PR that is the wpcc gem version. I use the gem git version until we have the final version (basically the PRs on wpcc that are waiting to be merged). When the final version of wpcc is released, I will change the git to the respective version.

Another gotcha is that today we have a syndicated version of the tool: https://partner-tools.moneyadviceservice.org.uk/workplace_pensions/index.html
As far as I understood the code, we don't need to do anything on frontend repo and removing from the frontend could not affect the syndication since it is serving via nginx static files. 

I opened another PR indicating that the old wpcc gem will no longer be used: https://github.com/moneyadviceservice/contribution_calculator/pull/7

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1795)


margOnline:
lgtm 👍

rnalexander:
:+1: :shipit:

PR #1799, opened 2017-08-29
[DEV APPROVED] Fix `bundle install` issues with latest `bundler`

igelstorm:
Currently, running `bundle install` for the first time with the latest version of `bundler` returns an error:

```
Downloading codeclimate-test-reporter-0.6.0 revealed dependencies not in the API or the lockfile (simplecov (< 1.0.0,
->= 0.7.1)).
Either installing with `--full-index` or running `bundle update codeclimate-test-reporter` should fix the problem.
```

There's a known workaround, which is to downgrade `bundler` to 1.14.6. Once the gems have been installed, `bundle install` works with no errors even with the latest `bundler` version, which is probably why this problem doesn't appear in Go builds or MAS dev environments.

The problem turned out to be an incorrectly specified dependency for `codeclimate-test-reporter` in the `Gemfile.lock` (similar problems are described [here](https://github.com/bundler/bundler/issues/5846), with suggestions that the cause is usually a corrupted lockfile).

To fix it, we locked `codeclimate-test-reporter` to the version currently used (since upgrading to the latest version would require code/configuration changes), and ran `bundle update codeclimate-test-reporter`, allowing bundler to correct the dependencies in the lockfile.

Since this fixes the problem, the workaround has been removed from the README.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1799)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,7 +89,7 @@ end
 group :test do
   gem 'capybara'
   gem 'chronic'
-  gem 'codeclimate-test-reporter', require: false
+  gem 'codeclimate-test-reporter', '0.6.0', require: false
   gem 'cucumber-rails', require: false
   gem 'database_cleaner'
   gem 'email_spec', '< 2' # DelayedJob integration removed in 2.0.0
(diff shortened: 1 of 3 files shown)


margOnline:
👍

margOnline:
Good description of the issue and the fix 😍

PR #1802, opened 2017-08-30
[DEV APPROVED] Update JavaScript dev dependencies

igelstorm:
This PR is very similar to https://github.com/moneyadviceservice/budget_planner/pull/484, and done for similar reasons.

The currently specified versions of PhantomJS and a number of Karma packages are very old (many years). These are only used to run specs, and the specs run fine with the most recent versions, so this updates them. `karma-sinon-chai` had to be locked to a less recent version, since more recent ones cause individual tests to fail.

Since these packages are only used in development, deploying this change to production should not cause any issues.

Without some of these updates, it's often not possible to run the specs at all on systems with relatively recent versions of PhantomJS (>=2) installed, which causes issues for development.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1802)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,26 +2,27 @@
   "name": "frontend",
   "private": true,
   "devDependencies": {
-    "chai": "^1.9.1",
-    "chai-jquery": "^1.2.3",
-    "karma": "^0.12.23",
+    "chai": "^3.5.0",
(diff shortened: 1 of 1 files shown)


tomas-stefano:
👍

margOnline:
👍

PR #1804, opened 2017-09-06
[DEV APPROVED] Moves the inclusion of JQuery higher in layout template

johnplayer1982:
# Moving JQuery higher in layout

There currently an issue with tracking and optimizely where JQuery is being loaded after both scripts, and optimizely requires JQuery, this is resulting in around a 40% of target users not seeing the A/B tests.

This PR moves the inclusion of JQuery to before the optimizely include tag.

## ASYNC

You'll notice that I have include an ``async`` attribute to the JQuery include tag, this will ensure that the script in non-blocking, and will allow the DOM content to be loaded while script is executed.  After a quick test I found the following performance impact:

| | DOMLOAD (Page is readable) | FULL LOAD (All images, scripts are complete) |
|-------|-------|-------|
| Current | 1.24s | 2s |
| ASYNC | 1.60s | 1.97s |

The DOM load is slightly impacted because the scripts are being executed at the same time as the DOM is being loaded.

While the full page load is slightly lower because the execution of jquery had already began earlier on in the loading of the page.

This will be tested in UAT and the impact on performance made clearer.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1804)

### DIFF
--- a/app/views/layouts/_base.html.erb
+++ b/app/views/layouts/_base.html.erb
@@ -9,6 +9,8 @@
 <head>
   <meta property="fb:pages" content="160564360681471" />
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+  <%= javascript_include_tag 'jquery/dist/jquery', async: 'async' %>
   <%= optimizely_include_tag if (Rails.env.production? || is_environment_on_uat?) %>
 
   <% if hide_elements_irrelevant_for_third_parties? %>
@@ -213,7 +215,7 @@
           j = d.createElement(s), dl = l != 'dataLayer' ? '&l=' + l : '';
       j.async = true;
       j.src = '//www.googletagmanager.com/gtm.js?id=' + i + dl;
-      f.parentNode.insertBefore(j, f);
+      f.parentNode.insertBefore(j, f.nextSibling);
     })(window, document, 'script', 'dataLayer', '<%= Rails.application.config.google_tag_manager_id %>');
   </script>
 <% end %>


rnalexander:
:+1:

willhall88:
can you also use the defer tag?

johnplayer1982:
Not in this case, the DEFER tag will wait until the page has finished loading before fetching and processing the JS, We would essentially end up with the same issue as we do currently

johnplayer1982:
Back into WIP as JQuery is delivered using requirejs and uploaded to our CDN, not as simple to just move a tag.

johnplayer1982:
@willhall88 @davidtrussler @rnalexander @tomas-stefano 

Sorry guys this needs a re-review, It wasn't as simple as moving JQuery out of requirejs and into a rails helper in the markup, as the application.js required jquery to be loaded through requirejs.

There has been no change to the load speed performance.  Although JQuery is called at the top of the page and also in requirejs, they are the same file, so by time the 1st instance is loaded, the 2nd instance in application.js will not need to do anything as jquery has already been fetched and processed.

davidtrussler:
Looks to me like it functions as expected - 👍

willhall88:
👍

norm:
`element.insertAdjacentHTML('afterend', ...)`?

tomas-stefano:
The insertAdjacentHTML is a good one, but I think would be more consistent ... since there is an insertBefore on js, I thought that there will be an insertAfter. But perhaps it is just me then :).

So I would leave up to @johnplayer1982 what is more readable:

`f.parentNode.insertBefore(j, f.nextSibling);`
or
`f.parentNode.insertAdjacentHTML('afterend', j);`

willhall88:
isn't there a .prepend() ?

willhall88:
https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore 😭

johnplayer1982:
Yep but .prepend(); is JQuery

willhall88:
ah, it's experimental still https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/prepend

PR #1805, opened 2017-09-11
[DEV APPROVED] 8556 - Sets a flexible width for syndicating debt advice locator tool (f2f)

johnplayer1982:
# Debt advice locator (f2f) tool width

Ticket https://moneyadviceservice.tpondemand.com/entity/8566

Currently when the debt advice locator (f2f) tool is mounted using the syndication script, the tool appears very narrow, around 300px.

This needs to be adjusted so that the tool expands to the width of its parent container (iframe needs to have 100% width).  This gives partners control over how wide the tool appears once mounted as it will always be the full width of the container they put it in.

| Before | After |
|-------|------|
|image|image|

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1805)


davidtrussler:
👍

PR #1809, opened 2017-09-19
[DEV APPROVED] 8577 - Refactor Error pages

Tr4pSt3R:
**Target Process ticket**
[Card: 8577](https://moneyadviceservice.tpondemand.com/entity/8577)

https://moneyadviceservice.tpondemand.com/entity/8577

**Summary**
This card picks up on the work done last week to push through some changes to the error pages. Some recommendations were made and it was suggested that a card should be made to complete these proposed refactoring. This card does not add new functionality but rather enhances the underlying code behind a feature that works already. 

**404 Page in English**

**500 Page in Welsh**

**QA / Testing**
You can test this by directly visiting the error pages for the HTML status codes for 404 and 500. 

404: http://localhost:3000/404?locale=en
500: http://localhost:3000/500?locale=cy

**Checklist**
+ [x] Tests passed.
+ [ ] Rubocop is passing for this PR (or any other lint like JShint). **not applicable**

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1809)

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -3,18 +3,17 @@ class ErrorsController < ApplicationController
   before_filter :fetch_exception
 
   def show
-    render :show, status: @status_code, formats: [:html]
+    render :show, status: params[:status_code], formats: [:html]
   end
 
   def display_skip_to_main_navigation?
-    @status_code != '404'
+    params[:status_code] != '404'
   end
 
   private
 
   def fetch_exception
     @exception = env['action_dispatch.exception']
-    @status_code = ActionDispatch::ExceptionWrapper.new(env, @exception).status_code
     @rescue_response = ActionDispatch::ExceptionWrapper.rescue_responses[@exception.class.name]
   end
 end
--- a/app/views/errors/show.html.erb
+++ b/app/views/errors/show.html.erb
@@ -4,40 +4,19 @@
 
 <p class="error-page__text<% if @rescue_response != :not_found %> error-page__text--only<% end %>"><%= t(".#{@rescue_response}.intro", default: t(".#{@rescue_response}.title")) %></p>
 
-<% if @rescue_response == :not_found %>
-  <p class="error-page__search-text"><%= t('.not_found.message') %></p>
+<p class="error-page__search-text"><%= t(".#{@rescue_response}.message") %></p>
 
-  <form data-dough-component="ClearInput" action="/en/search" method="get" role="search" class="error-page__search search">
-    <label class="visually-hidden" for="search"><%= t('search_box.label') %></label>
-    <input data-dough-clear-input class="search__input" id="search" required type="search" name="query" autocomplete="off" placeholder="<%= t('.not_found.search_placeholder') %>" />
+<form data-dough-component="ClearInput" action="/en/search" method="get" role="search" class="error-page__search search">
+  <label class="visually-hidden" for="search"><%= t('search_box.label') %></label>
+  <input data-dough-clear-input class="search__input" id="search" required type="search" name="query" autocomplete="off" placeholder="<%= t(".#{@rescue_response}.search_placeholder") %>" />
 
-    <button data-dough-clear-input-button type="reset" class="search__clear">
-      <span class="icon icon--clear"></span>
-      <span class="visually-hidden"><%= t('search_box.clear_hidden_text') %></span>
-    </button>
+  <button data-dough-clear-input-button type="reset" class="search__clear">
+    <span class="icon icon--clear"></span>
+    <span class="visually-hidden"><%= t('search_box.clear_hidden_text') %></span>
+  </button>
 
-    <button class="search__submit" type="submit">
-      <span class="icon icon--search"></span>
-      <span class="visually-hidden"><%= t('search_box.button_text') %></span>
-    </button>
-  </form>
-<% end %>
-
-<% if @rescue_response == :internal_server_error %>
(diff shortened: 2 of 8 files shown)


app/views/errors/show.html.erb
```diff
@@ -8,7 +8,7 @@
 
 <form data-dough-component="ClearInput" action="/en/search" method="get" role="search" class="error-page__search search">
   <label class="visually-hidden" for="search"><%= t('search_box.label') %></label>
-  <input data-dough-clear-input class="search__input" id="search" required type="search" name="query" autocomplete="off" placeholner="<%= t(".#{@rescue_response}.search_placeholder") %>" />
+  <input data-dough-clear-input class="search__input" id="search" required type="search" name="query" autocomplete="off" placeholder="<%= t(".#{@rescue_response}.search_placeholder") %>" />
```
Tr4pSt3R: @pmcostello ♻️

app/controllers/errors_controller.rb
```diff
@@ -13,7 +13,11 @@ def display_skip_to_main_navigation?
   private
 
   def fetch_exception
-    @exception = env['action_dispatch.exception']
+    @exception = if env['action_dispatch.request.path_parameters'][:status_code] == '404'
+       ActionController::RoutingError.new(:page_not_found)
```
tomas-stefano: Why the controller fetch an exception to raise another?

features/step_definitions/error_page_steps.rb
```diff
@@ -0,0 +1,15 @@
+Given(/^that I visit the internal server error page in my "([^"]*)"$/) do |language|
+  locale = language_to_locale(language)
+
+  internal_server_error_page.load(locale: locale)
+end
+
+Given(/^that I visit a non\-existent page in my "([^"]*)"$/) do |language|
+  locale = language_to_locale(language)
+
+  page_not_found_page.load(locale: locale)
```
tomas-stefano: If this is the step to visit a non-existent page why is visiting the "404" page which exists on our app? Why not trying to access another like "/this-page-does-not-exist" url?

app/controllers/errors_controller.rb
Tr4pSt3R: @tomas-stefano for some reason `env['action_dispatch.exception']` or `request. env['action_dispatch.exception']` always returns a `nil` value. This means by default every error is forced to render as an _internal server error_ issue even when it's not a `500` error. So one thing is to catch the error early before it's coerced by default to be a `500` error even when it isn't.

Do you see what I mean. please?

features/step_definitions/error_page_steps.rb
Tr4pSt3R: @tomas-stefano there's a spec for it already. https://github.com/moneyadviceservice/frontend/blob/master/spec/requests/404_spec.rb#L12

This test is to examine the message we get back from the rendered 404/500 page itself.

tomas-stefano: But what user goes direct to the 404 page? Shall we not mimic the user behaviour on our cucumber specs?

app/controllers/errors_controller.rb
tomas-stefano: So is the message "This means by default every error is forced to render as an internal server error issue even when it's not a 500 error"  sounds that is fixing a bug then?

Tr4pSt3R: It is a bug. There was some work done to remove the Gaffe gem() about two years ago. This work may have accidentally introduced the bug. 

Here is a bug-fix by @igelstorm on this very issue from about a month ago (https://github.com/moneyadviceservice/frontend/pull/1789).

tomas-stefano: I didn't understand. Is this PR fixing a bug or refactor existing feature?

features/support/env.rb
```diff
@@ -56,6 +56,10 @@
   create(:corporate_partner)
 end
 
+Before('@allow_rescue') do
+  ActionController::Base.allow_rescue = true
+end
+
```
tomas-stefano: Why this line is here? This line means in all tests we allow exceptions. Is there what is intended to do it?

Tr4pSt3R: @tomas-stefano this has been put there so that exceptions are treated in the same way as it will be treated on production. Otherwise it will dump a stack trace of the error rather than redirecting it to the ErrorController to ensure that the 404 page gets picked up.

Which is what we needed.

willhall88: This will allow exceptions to be rescued just for tests that are tagged with `@allow_rescue` 
https://github.com/moneyadviceservice/frontend/pull/1809/files#diff-e76069191d84db196d4a91586c5b8ad9R19

app/controllers/errors_controller.rb
margOnline: Looks like its refactoring an existing feature as well as addressing tech debt

margOnline:
nice work 👍

willhall88:
The code itself looks good. 
But is the whole `fetch_exception` method needed, does the status_code not get passed through to the controller as a variable already?

Tr4pSt3R:
@willhall88 I wondered myself why that was needed but it's a remnant from how the application was set up to catch failures/exceptions. I don't want to change much as it may affect other controllers. 

`@status_code` does not get passed to the **ErrorController** by default. It has to be deduced.

willhall88:
👍  code looks good! just make sure to rebase for commit history

PR #1811, opened 2017-09-26
[DEV APPROVED] 8549 - Adds links for tools on mobile nav

thecountgs:
# The Global Nav does not surface tool links on mobile

TP [8549](https://moneyadviceservice.tpondemand.com/entity/8549)

a) These tools are often interacted with on Desktop so it would be valuable to surface them on mobile as well.

b) Fix vertical scrolling issue introduced to mobile nav when tools are displayed.

**Bonus**
Fix body scrolling issue discovered while testing mobile nav.

## Solution:

Change tools in mobile nav visibility to `display: block;`
Add `'no-scroll` class to body when mobile nav is opened.
Update subnav view  by adding `is-hidden` class and remove class with javascript when subnav is triggered.
Increase bottom padding in `.global-subnav` to allow the close button to be displayed.

## Notes:
This work led to the discovery of some limitations to our Karma testing environment. 

**Specifically:**
Karma is not set up to compile SCSS and include these stylsheets in its fixtures. This means that javascript which modifies CSS properties can not be directly tested.

screenshot_20170926-161115

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1811)

### DIFF
--- a/app/assets/javascripts/components/GlobalNav.js
+++ b/app/assets/javascripts/components/GlobalNav.js
@@ -233,7 +233,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities', 'common'], 
 
     this.$globalSubNavHeading.click(function(e) {
       e.preventDefault();
-      self._toggleMobileSubNav(this);
+      self._returnMobileNav(this);
     });
 
     this.$mobileNavClose.click(function(e){
@@ -304,6 +304,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities', 'common'], 
   GlobalNav.prototype._toggleMobileNav = function() {
     this.$globalNav.toggleClass('is-active');
     this.$mobileNavOverlay.toggleClass('is-active');
+    $('body').addClass('no-scroll');
 
     // If we just closed the nav, reset all subnavs
     if (!this.$globalNav.hasClass('is-active')) {
@@ -311,6 +312,7 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities', 'common'], 
       this.$mobileNavOverlay.removeClass('is-active');
       this.$globalNavClumps.removeClass('is-active');
       this.$globalNavClump.removeClass('is-active');
+      $('body').removeClass('no-scroll');
     }
   };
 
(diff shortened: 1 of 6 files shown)


spec/javascripts/tests/GlobalNav_spec.js
```diff
@@ -1,4 +1,4 @@
-describe('GlobalNav', function() {
+describe.only('GlobalNav', function() {
```
davidtrussler: Need to remove .only here!

app/assets/javascripts/components/GlobalNav.js
```diff
@@ -304,20 +304,35 @@ define(['jquery', 'DoughBaseComponent', 'mediaQueries', 'utilities', 'common'],
...
+  };
+
+  GlobalNav.prototype._returnMobileNav = function(index) {
     $(index)
       .parents('[data-dough-nav-clump]').toggleClass('is-active')
       .parents('[data-dough-nav-clumps]').toggleClass('is-active');
+      setTimeout(function(){
+        $('[data-dough-subnav]').addClass('is-hidden');
+      }, 400);
   };
```
davidtrussler: I wondered if this could be added in the CSS as a transition-delay property. The animation of the nav is done there so would make sense to keep it together and would share the variable for the time delay.

app/assets/stylesheets/components/common/_global_subnav.scss
```diff
@@ -152,11 +156,10 @@ $global-nav-transition-duration: 300ms !default; // Defined in _global_nav.scss
 }
 
 .global-subnav__tool-category {
-  display: none;
+  display: table-cell;
```
davidtrussler: Not sure about table-cell here. We have used flexbox on desktop and table-cell as a fallback on non-compliant browsers. Not a big deal maybe: depends on what, erm, design we're working to I suppose.

app/assets/javascripts/components/GlobalNav.js
thecountgs: Per our conversation: `transition-delay` can't be applied to the `display` property.

app/assets/stylesheets/components/common/_global_subnav.scss
thecountgs: Updated per our conversation.

spec/javascripts/tests/GlobalNav_spec.js
thecountgs: Done!

davidtrussler:
👍  - looks good!

johnplayer1982:
:+1:

PR #1813, opened 2017-09-27
8543 - Update budget planner to 5.0.3.771

tomas-stefano:
TP: https://moneyadviceservice.tpondemand.com/entity/8543

So this PR bumps the budget planner version which contains the budget planner referral url feature and adds the referral url to the database. The whole work you can see here: https://github.com/moneyadviceservice/budget_planner/pull/485

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1813)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170719110635) do
+ActiveRecord::Schema.define(version: 20170914164504) do
 
   create_table "action_plans_expense_items", force: :cascade do |t|
     t.string  "kind",       limit: 256,             null: false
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -129,6 +129,7 @@
     t.integer  "user_id",            limit: 4
     t.string   "commited_from",      limit: 255
     t.string   "last_commited_from", limit: 255
+    t.text     "referral_url",       limit: 65535
```
margOnline: I thought you had removed the limit due to the length issues with different versions of mysql?

tomas-stefano: I had problem with to add an index on the text field but the text field length does not raise any problems. Since I end up not adding an index, then everything is fine! :)

willhall88:
👍

margOnline:
👍

PR #1830, opened 2017-10-26
Bump cost calculator version

tomas-stefano:
TP: https://moneyadviceservice.tpondemand.com/entity/8602

So this PR bumps the cost calculator builder version, which makes possible to add a image to the cost calculator builder tools.

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1830)


tomas-stefano:
This PR was already thumbed and reverted it when we had the db problems in production. Since tonight we will do the deployment to fix this db problems, I will merge it. Thanks for the comprehension.

PR #1888, opened 2018-01-22
[DEV APPROVED] - 8665 Refactor category and ancillary pages to work with CMS gem

gotbadger:
**Target Process ticket**

[TP 8665](https://moneyadviceservice.tpondemand.com/entity/8665-use-mas-cms-client-gem-on)

**Summary**
This PR involves removing the use of the `Core` category classes and using `Mas::Cms` replacement where possible.

The scope of removing all `Core` category usage is quite large so this PR does not completely remove all references but progresses the code towards this focusing on the controller pages as an initial effort.

When the global nav is updated as part of [TP 8667](https://moneyadviceservice.tpondemand.com/entity/8667-use-mas-cms-client-gem-on) we should look at removing more things especially some of the complicated navigation/category methods in [app/application_controller.rb](https://github.com/moneyadviceservice/frontend/blob/75fe5ece7ab55051e23240b3f8b760510338015c/app/controllers/application_controller.rb#L126-L161)

**QA / Testing**
Manual testing should check for regressions on the following areas:
 - article pages
 - category pages
 - corporate category pages
 - corporate article pages

**Checklist**

+ [x] Tests passed.
+ [ ] Commit history reviewed (clear commit messages).

---
This change is [Reviewable](https://reviewable.io/reviews/moneyadviceservice/frontend/1888)


gotbadger:
I cant view the codeclimate issue on this one so if anyone can enlighten me I can fix it 😉

tomas-stefano:
@gotbadger Code Climate should be on the Keepass. Did you have acess to the Keepass?

app/controllers/corporate_categories_controller.rb
```diff
@@ -20,15 +19,12 @@ def paywall_debt_advice_evaluation_toolkit
     end
   end
 
-  def article_interactor
-    Core::CorporateReader.new(default_article_id)
+  def article_resource
+    Mas::Cms::Article.find(default_article_id, locale: I18n.locale)
```
tomas-stefano: Not sure if this was intended but before this PR, the code was trying to find Corporate pages (using Core::CorporateReader) and now it is replaced for Article pages. Is that what is intended? I am asking because for corporate article there is on the gem the class `Mas::Cms::Corporate`.

app/decorators/corporate_category_content_decorator.rb
```diff
@@ -1,11 +1,12 @@
...
-    when Core::Article
+    when Core::Article, Mas::Cms::Article
       h.send("#{object.class.to_s.demodulize.underscore}_path", object.id, locale: I18n.locale)
-    when Core::Category
+    when Core::Category, Mas::Cms::Category
       h.send("corporate_#{object.class.to_s.demodulize.underscore}_path", object.id, locale: I18n.locale)
-    when Core::Other, Core::Video
+    when Core::Other, Core::Video, Mas::Cms::Other, Mas::Cms::Video
       "/#{I18n.locale}/#{object.type.pluralize}/#{object.id}"
     end
```
tomas-stefano: Some considerations:

1) IMHO, the best course of action is not adding `TODO` on the code as we have the history on not looking again and leave the code as it is for a long time. Make sure that it is not technical debt and if is inevitable then make sure that there is a card that contains the issue.
2) I am not sure if I missing something but I don't we are using `Mas::Cms::Other` or `Mas::Cms::Video`. Are we using this classes? I would delete this mentions on the code if we are not fetching yet.
3) Since we are already replacing the corporate with the new code why not to delete the Core::Article mention and Core::Category mention?

app/decorators/category_content_decorator.rb
```diff
@@ -15,11 +15,12 @@ def label
...
+  # TODO: Remove Core type checks when they have been removed
   def path
     case object
-    when Core::Article, Core::Category
+    when Core::Article, Core::Category, Mas::Cms::Article, Mas::Cms::Category
       h.send("#{object.class.to_s.demodulize.underscore}_path", object.id, locale: I18n.locale)
-    when Core::Other, Core::Video
+    when Core::Other, Core::Video, Mas::Cms::Other, Mas::Cms::Video
       "/#{I18n.locale}/#{object.type.pluralize}/#{object.id}"
     end
```
tomas-stefano: Same consideration that I wrote below.

app/decorators/corporate_category_content_decorator.rb
gotbadger: Yeah ok we can probably do this now 👍

gotbadger: @tomas-stefano for (3) I looked into this a bit more and I think we should leave `Core::Article` and `Core::Article` there are still quite a few things like ActionPlan that inherit from Article and have not yet been updated and potentially have category stuff in so  I would feel more comfortable addressing this once those controllers have been migrated

spec/decorators/category_content_decorator_spec.rb
```diff
@@ -91,6 +91,7 @@
       let(:item) { build :action_plan }
 
       it 'calls the correct path helper' do
+        binding.pry
```
willhall88: Please remove

gotbadger: 😱 oops!

spec/controllers/categories_controller_spec.rb
```diff
@@ -1,71 +1,30 @@
...
       end
 
       it 'assigns @breadcrumbs to the result of breadcrumb reader' do
-        get :show, id: category.id, locale: I18n.locale
-
         expect(assigns(:breadcrumbs)).to eq(breadcrumbs)
       end
     end
-
-    context 'when the category does not exist' do
```
KateWilkinson: Just curious - have these routing error/redirect tests been removed because this stuff is all handled in the ApplicationController now? Do you know if this is being tested elsewhere?

gotbadger: yeah these should probably be tested in the application controller

gotbadger: This was done in 7e25924

app/domain/breadcrumb_trail.rb
```diff
@@ -1,29 +1,38 @@
...
-          [HomeCategory.new]
-        end
-      when ToolCategory
-        RootToNodePath.build(item, category_tree)
-      when Core::StaticPage
+  def self.build(item, category_tree = nil)
+    case item
+    when Mas::Cms::Category
+      # recurse and find categories instead of passing in category_tree?
+      if item.parent_id.present?
```
willhall88: could we move this if block into its own method? It will keep the case statement a little cleaner, wdyt?

app/domain/breadcrumb_trail.rb
```diff
@@ -1,29 +1,38 @@
...
+        [HomeCategory.new]
+      end
+    when ToolCategory
+      RootToNodePath.build(item, category_tree)
+    when Core::StaticPage
+      [HomeCategory.new]
+    when Core::NewsArticle
+      [HomeCategory.new, NewsCategory.new]
+    else
+      if item.categories.empty?
```
willhall88: as above, perhaps move this if block into it's own smaller method

app/domain/breadcrumb_trail.rb
```diff
@@ -1,29 +1,38 @@
...
-        end
+        []
+      end
+    end.map do |element|
+      Breadcrumb.new(element)
     end
+  end
 
-    categories.map(&Breadcrumb.public_method(:new))
+  def self.build_trail(category_id, children = [])
```
willhall88: :+1: nice

spec/controllers/application_controller_spec.rb
```diff
@@ -68,4 +68,34 @@ def index
       expect(subject.show_floating_chat?).to be(false)
     end
   end
+
+  context 'when cms resouce not found' do
```
willhall88: `resource` *

spec/controllers/application_controller_spec.rb
```diff
@@ -68,4 +68,34 @@ def index
...
+      def index
+        cms_response = Faraday::Response.new.finish(
+          status: 302,
+          response_headers: { location: '/test' }
+        )
+        raise Mas::Cms::HttpRedirect, cms_response
+      end
+    end
+
+    it 'rediects to given url' do
```
willhall88: `redirects` *

app/domain/breadcrumb_trail.rb
gotbadger: Sorting out `BreadcrumbTrail` anymore than I have done is probably a bit of a rabbit hole in-fact I almost reverted this change out. I think probably the best time to attempt to sort out this class would be when/if we remove the category methods in application controller.

In terms of moving this to a method right now I wanted to keep it somewhat similar to the original version till we can take on the whole class. Likely when we do so this where clause will reduce significantly (or even go completely) and be less of an issue.

willhall88: My only concern to that would be the 

>  when/if we remove the category methods in application controller

*if* we don't end up doing that work in the application controller, then it's more than likely this work won't get done and this code is left, in my eyes, as tech debt. 

When I currently look at the switch statements, I find it quite hard to follow and think it would be slightly easier if the `if` blocks were moved down the class into their own small methods.

I haven't looked at the current backlog of tickets around this, so my question would be, Do you see this work being done as part of another ticket in the very near future?

PR #1892, opened 2018-01-26
Homepage Beta Search Component

thecountgs:
This PR creates the search component for the MAS  beta homepage. 

[TP](https://moneyadviceservice.tpondemand.com/entity/8680-homepage-create-central-search-box)

## Description
As a new user I want to start my journey on the Money Advice Service site using site search but I don't always see the top right search box.
 
**What needs to happen:**
Implement a UI element below the leader image that features site search, centred on the page.

|English | Welsh|
|-------|--------|
| screen shot 2018-01-26 at 11 08 27| screen shot 2018-01-26 at 11 08 53|

### DIFF
--- a/app/assets/stylesheets/components/page_specific/home_beta/_central_search.scss
+++ b/app/assets/stylesheets/components/page_specific/home_beta/_central_search.scss
@@ -1,12 +1,12 @@
 .central_search-beta {
-	@extend %clearfix;
-	background-color: $color-green-primary;
+  @extend %clearfix;
+  background-color: $color-green-primary;
   margin-top: $baseline-unit*2;
   width:100%;
 }
 
 .central_search-beta__content {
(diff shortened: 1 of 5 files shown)


davidtrussler:
A couple of general thoughts, as these are not entirely clear from the design supplied: 

1) Should the component fit the full screen width on wider viewports, as the header and footer currently do? If not then I think it should fit to the other components where it is now a little wider. 

image

2) Do we need this component on width of display where the mobile search box is displayed? Looks like unnecessary duplication IMO. 

image

thecountgs:
Good point @davidtrussler I modified the view so the search component now goes full width. Thanks to @johnplayer1982 for the assist.

davidtrussler:
👍

gotbadger:
👍

PR #1896, opened 2018-01-31
[DEV APPROVED] - 8865 remove old news article code

gotbadger:
[TP 8865](https://moneyadviceservice.tpondemand.com/entity/8865-remove-unused-news-article-code)

**Summary**
News articles are no longer used in frontend with `/en/news` being a redirect in production to a [standalone blog](https://www.moneyadviceservice.org.uk/blog/). 

This code hanging around was incurring technical debt making the changes to add `mas-cms-client` to frontend more difficult that they should be.

This PR removes all the unused code and removes references to news articles from a few other tests.


willhall88:
does this file also need to go? 
https://github.com/moneyadviceservice/frontend/blob/master/public/news-sitemap.xml

gotbadger:
@willhall88 maybe. Its still served from https://www.moneyadviceservice.org.uk/news-sitemap.xml but looks like it has contents that suggested it could go. Not knowing the code base that well I've tried to be fairly conservative in what I've removed, taking out files that i know are definitely not needed.

If you think this could go without any issue it can be removed. WDYT?

willhall88:
I probably have as much knowledge as you to be honest! It was definitely a question of whether it was now redundant, rather than a statement. If it looks like we still need it then, I'm happy for it to remain in the codebase.

tomas-stefano:
@gotbadger can you talk with Ed and raise that https://www.moneyadviceservice.org.uk/news-sitemap.xml and it is probably/potentially not used anymore, please?

PR #1900, opened 2018-02-02
[DEV APPROVED] 8675 homepage leader image

davidtrussler:
**Target Process ticket**
[TP8675](https://moneyadviceservice.tpondemand.com/RestUI/Board.aspx#page=userstory/8781&appConfig=eyJhY2lkIjoiREI4ODcwRjkxMDNDRTM2NTlBMzhDNTRBRTVBNUU1N0UifQ==&searchPopup=userstory/8675)

**Summary**
This ticket adds the new leader image to the homepage. The mobile view serves up one of three images on a random basis when the page is loaded. The text is added via SVG to retain clarity in all views when scaled. 

**Examples**

***Mobile views***

image

image

image

***Desktop view***

image


willhall88:
Please could you tidy up the commit history

davidtrussler:
Have tidied up commit history though my personal preference is to do this after PR approval.

app/helpers/home_helper.rb
```diff
@@ -12,6 +12,12 @@ def markdown(text)
     remove_p_tags_from(document).html_safe
   end
 
+  def leader_text(name, size, x, y, rotation)
```
Tr4pSt3R: Hey @davidtrussler can we pair on this to add a BE spec test for this helper please?

davidtrussler: We can indeed (though this isn't the highest priority right now though to be honest) but would be well worth it at some point.

davidtrussler: @Tr4pSt3R Thinking about it, I'd like to get this PR merged into the feature branch. Can we look at adding the test after that?

Tr4pSt3R: @davidtrussler I'll work on adding the test to it now and submit a PR to this branch

Tr4pSt3R: After having a chat withth @margOnline && @davidtrussler we will leave the testing aspect of this for now since it's no longer the priority. FinCap is now the priority. You can merge this into the feature branch and make a note to add test to this in the future :(

PR #1901, opened 2018-02-06
[DEV APPROVED] Creates feature spec for the show_all? flag on money manager/universal credit

scott-ad-riley:
[TP Link](https://moneyadviceservice.tpondemand.com/RestUI/Board.aspx#page=board/4771110243071511547&appConfig=eyJhY2lkIjoiRjk1Qzc3Njg4REFDNTUyNTBGODhGNTE5MEM5M0JCN0UiLCJhcHBDb250ZXh0Ijp7InByb2plY3RDb250ZXh0Ijp7Im5vIjpmYWxzZX0sInRlYW1Db250ZXh0Ijp7Im5vIjpmYWxzZX19fQ==)

**Summary**
This PR creates a feature spec for the TP Link card above. The logic is implemented in https://github.com/moneyadviceservice/universal_credit/pull/216.

**Checklist**

+ [ ] Tests passed.
+ [ ] Rubocop is passing for this PR (or any other lint like JShint).
+ [ ] Code Climate passed for this PR.
+ [ ] Commit history reviewed (clear commit messages).

**N.B.** The test is failing because the universal credit PR has not been merged and so we cannot bump the version of UC yet. Once that has been merged we will add that bump to this PR and the feature spec should pass.

### DIFF
--- a/features/money_manager.feature
+++ b/features/money_manager.feature
@@ -20,7 +20,7 @@ Scenario: Signing in after completing questionnaire saves answers
   And   I select a country
   And   I sign in
   Then  I get my results
-  
+
 Scenario: Signing in should overwrite old answers with new ones
   Given I am signed in
   And   I am on the Money Manager tool
@@ -30,3 +30,10 @@ Scenario: Signing in should overwrite old answers with new ones
   And   I answer the questions differently
   And   I sign in
   Then  I have the latest answers
+
+Scenario: Passing a special parameter in the url should show all content
+  Given I am signed in
+  And   I am on the Money Manager tool
+  And   I complete the series of questions
+  And   I visit the show all url
+  Then  I see all of the advice
--- a/features/money_manager.feature
+++ b/features/money_manager.feature
@@ -31,6 +31,7 @@ Scenario: Signing in should overwrite old answers with new ones
   And   I sign in
   Then  I have the latest answers
 
+@no-javascript
 Scenario: Passing a special parameter in the url should show all content
   Given I am signed in
   And   I am on the Money Manager tool
--- a/features/step_definitions/money_manager_steps.rb
+++ b/features/step_definitions/money_manager_steps.rb
@@ -29,6 +29,10 @@ Given(/^I get my results$/) do
   expect(page.current_path).to eql('/en/tools/money-manager/to-read')
 end
 
+When(/^I visit the show all url$/) do
+  money_manager_show_all_advice_page.load
+end
+
 Then(/^I should not see any hint of my details when I re-visit the tool$/) do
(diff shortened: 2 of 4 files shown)


features/money_manager.feature
```diff
@@ -29,4 +29,11 @@ Scenario: Signing in should overwrite old answers with new ones
...
   And   I sign in
-  Then  I have the latest answers
\ No newline at end of file
+  Then  I have the latest answers
+
+Scenario: Passing a special parameter in the url should show all content
+  Given I am signed in
+  And   I am on the Money Manager tool
+  And   I complete the series of questions
+  And   I visit the show all url
```
Tr4pSt3R: Hi @scott-ad-riley I suspect the `Show All` will be a button to be clicked right? In which case rather than following a URL in this manner why not have this step instead `And I click the "Show all" button to open all advice tabs`.

scott-ad-riley: There is no button - this is an admin-only feature (without having to implement an admin and login). The people who are admins/content creators already know about the param for the url. 

This PR and the implementation one in universal credit are for upgrading the existing feature of the `?show_all` in the url - they aren't actually implementing that feature.

Tr4pSt3R: @scott-ad-riley okay I get what you mean

features/money_manager.feature
```diff
@@ -29,4 +29,11 @@ Scenario: Signing in should overwrite old answers with new ones
   When  I am on the Money Manager tool
   And   I answer the questions differently
   And   I sign in
-  Then  I have the latest answers
\ No newline at end of file
+  Then  I have the latest answers
+
+Scenario: Passing a special parameter in the url should show all content
```
Tr4pSt3R: Hi @scott-ad-riley is it worth tagging this scenario to explicitly communicate the fact that it's designed to work with/without JS?

```
@no-javascript
Scenario: Passing a special parameter in the url should show all content
.
.
```

scott-ad-riley:
Hi @Tr4pSt3R 

Please see the PR description:
>N.B. The test is failing because the universal credit PR has not been merged and so we cannot bump the version of UC yet. Once that has been merged we will add that bump to this pR and the feature spec should pass.

features/money_manager.feature
scott-ad-riley: I'll add it in.

Tr4pSt3R:
Sorry @scott-ad-riley. I just read the PR desc. Yup that makes sense.

scott-ad-riley:
Great - I'll squash once I can get the UC PR in and then I can bump the version of this. Not sure how/why CI is passing with the failing spec?

PR #1907, opened 2018-02-12
[DEV APPROVED] - 8667 Update site to use new clumps/footer methods

gotbadger:
**Target Process ticket**

[TP 8667](https://moneyadviceservice.tpondemand.com/entity/8667-use-mas-cms-client-gem-on)

**Summary**
This PR updates frontend to use the `mas-cms-client` gem to deal with global navigation elements in both the header (aka clumps) and footer (contact panels).

**QA / Testing**
We should test that the global nav and footer are identical to the production site


app/controllers/application_controller.rb
```diff
@@ -17,8 +17,7 @@ class ApplicationController < ActionController::Base
 
   before_action :fetch_footer_content
   def fetch_footer_content
-    interactor = Core::FooterReader.new('footer')
-    @footer = interactor.call
+    @footer = Mas::Cms::Footer.find('footer')
```
tomas-stefano: Nice!! 😍

PR #1915, opened 2018-03-02
[DEV APPROVED] update robots.txt file to add User-agent: *

pmcostello:
simple change to robots.txt file to fix 

The Disallow statements are currently being ignored by bots because of the missing User-Agent: * declaration.

### DIFF
--- a/app/views/sitemap/robots.text.erb
+++ b/app/views/sitemap/robots.text.erb
@@ -1,3 +1,4 @@
+User-agent: *
 Sitemap: <%= @base_url %>/sitemap.xml
 Disallow: /en/campaigns/free-debt-advice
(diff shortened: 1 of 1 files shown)


app/views/sitemap/robots.text.erb
```diff
@@ -1,3 +1,4 @@
+User-agent: *
```
Tr4pSt3R: @pmcostello is this to block *all* crawlers from meandering through the Frontend site as per https://support.google.com/webmasters/answer/1061943?hl=en

pmcostello: yeah, the default for the robots file is to allow * crawler bots but unless the User-Agent is explicitly declared the Disallow directives are being ignored

PR #1918, opened 2018-03-15
Add cache to the development mode using an env var

tomas-stefano:
**No TP cards** - but it is something that I did to help build the [8974](https://moneyadviceservice.tpondemand.com/entity/8974-land-transaction-tax-calculator-stamp-duty)

So there are a lot of times which we are not developing anything that integrates
with external APIS and since the development environment does not have
cache the **page load is slow every single page**. This commit fixes that
by adding the possibility to enable cache in development every time is
necessary.

Obs.: I added the .env-example the cache false as default because
the cache should be enabled intentionally by everyone that is running
the application.

And this is the development mode so does not affect production at all.


johnplayer1982:
:+1: love it!

README.md
```diff
@@ -64,6 +64,16 @@ To start the application:
 $ foreman s
 ```
 
+The site does a lot of requests to the CMS application, so in case
+**you are not developing anything that integrates with CMS**, you can
+enable the application cache in development mode by running:
```
margOnline:
```
The site makes a lot of requests to the CMS application. If
**you are not developing anything that integrates with CMS**, you can
enable the application cache in development mode by running:
```

tomas-stefano: ♻️

PR #1922, opened 2018-03-28
Feature/change site search

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/8785-site-search-switchover)

## Context

Google CSE is due to expire on the 31st March 2018
 
Needs to be replaced with an alternative before the CSE license expires.
 
Chosen replacement for Google CSE
https://www.algolia.com

## Implementation

So the implementation replaces the existing Google code and uses the site search gem.


app/decorators/search_result_collection_decorator.rb
```diff
@@ -10,19 +10,7 @@ def context
...
-    end
-  end
-
-  def total_results
-    if object.total_results > result_limit
-      result_limit
-    else
-      object.total_results
-    end
+    object.page + 1
```
willhall88: presuming this is down to zero indexing?

features/search.feature
```diff
@@ -11,41 +11,3 @@ Feature: Searching
     Given I am on the home page
     When I search for something relevant
     Then I should see search results
-
-  Scenario: A relevant message is displayed when no results are found
-    Given I am on the home page
-    When I search for something irrelevant
-    Then I should see no search results
-    And I should prompted to try another search term
```
willhall88: what is the expectation of this scenario now?

features/search.feature
```diff
@@ -11,41 +11,3 @@ Feature: Searching
...
-    Given I am on the home page
-    When I search for an incorrectly spelt term with no results
-    Then I should have corrected search results
-
-  Scenario: A relevant message is displayed if a search is performed with no keywords
-    Given I am on the home page
-    When I submit a search with no query
-    Then I should see the search page
-
-  Scenario: Browsing paginated results
```
willhall88: should this scenario not be kept?

features/support/vcr.rb
```diff
@@ -37,4 +30,6 @@
   c.filter_sensitive_data('<GOOGLE_API_KEY>') { ENV['GOOGLE_API_KEY'] }
   c.filter_sensitive_data('<GOOGLE_API_CX_EN>') { ENV['GOOGLE_API_CX_EN'] }
   c.filter_sensitive_data('<GOOGLE_API_CX_CY>') { ENV['GOOGLE_API_CX_CY'] }
```
willhall88: can these google keys now be deleted?

spec/decorators/search_result_collection_decorator_spec.rb
```diff
@@ -1,212 +1,137 @@
...
-      context 'when the current page is 5' do
-        before { result_collection.page = 5 }
+      context 'when fifth page' do
+        let(:page) { 4 }
+
         it { is_expected.to be_falsey }
       end
     end
 
     describe '#last_page?' do
```
willhall88: given we have deleted the feature for last page and no longer limited to only 5 pages are these tests still relevant?

spec/spec_helper.rb
```diff
@@ -63,6 +57,8 @@
   c.filter_sensitive_data('<GOOGLE_API_KEY>') { ENV['GOOGLE_API_KEY'] }
   c.filter_sensitive_data('<GOOGLE_API_CX_EN>') { ENV['GOOGLE_API_CX_EN'] }
   c.filter_sensitive_data('<GOOGLE_API_CX_CY>') { ENV['GOOGLE_API_CX_CY'] }
```
willhall88: same question. can these now be removed?

app/decorators/search_result_collection_decorator.rb
tomas-stefano: Yes.

features/search.feature
tomas-stefano: There is no spelling suggestion out of the box on Algolia. It is a feature that has been deleted.

spec/decorators/search_result_collection_decorator_spec.rb
tomas-stefano: Yes because if the last page is 10 for example it should return if I am on the last page and should not display the next button.

spec/spec_helper.rb
tomas-stefano: Yes. Updated ♻️

features/support/vcr.rb
tomas-stefano: Yes. Updated ♻️

tomas-stefano:
@margOnline thank you. Yes they were deliberated deleted for two reasons. First the spelling suggestion was kind trick to implement on time constraint that we had so I remove all the feature. Second: the browsing pagination features were totally Google based so they kind need of a rewrite to be more flexible. I am planning to implement this in today and open a separate PR today with the new tests. I push this explanation now on the commit message.

features/search.feature
tomas-stefano: This browsing pagination features were totally Google based so they kind need of a rewrite to be more flexible. I will open a separate PR today containing the new feature tests for browing and pagination.

features/search.feature
willhall88: ah, so this was based on spelling suggestions not just a generic 'no results found' message then?

spec/decorators/search_result_collection_decorator_spec.rb
willhall88: ok so taking your comment above I'm happy for another PR to be opened for a feature test around this then :)

willhall88:
:+1: happy to approve this PR to get the work finished. Looking forward to seeing the PR with the feature tests to cover the new behaviour :)

PR #1933, opened 2018-04-12
[DEV APPROVED] Add feature tests for site search including pagination

tomas-stefano:
**Target Process ticket**

[TP card](https://moneyadviceservice.tpondemand.com/entity/9100-frontend-site-search-pagination-feature-tests)

**Summary**

The feature tests on the frontend codebase which checked how pagination worked were too tightly tied to Google site search and couldn't be used.  We need to write replacement feature tests which cover pagination as it works under Algolia (and hopefully any future search engines we may use.)
 
This card will be finished when there are sufficient feature tests to provide confidence in that section of the code base moving forward.

**Checklist**

+ [x] Tests passed.

### DIFF
--- a/features/search.feature
+++ b/features/search.feature
@@ -1,13 +1,51 @@
-Feature: Searching
+Feature: Searching on MAS site
   As a user visiting the website
   I want to search by keyword
   So that I can find the content I am looking for
 
-  Scenario: The search box is visible in the header
-    When I visit the home page
-    Then I should see the search box
-
-  Scenario: Search can be performed
+  Background:
     Given I am on the home page
-    When I search for something relevant
-    Then I should see search results
+
+  Scenario: Search on first result page
+    When I search the query "money"
+    Then I should see "10" results
+    And I should see the search results
+      | title                                             |
+      | Money Manager tool for Universal Credit claimants |
+      | Money Advice Toolkits                             |
+      | Our board                                         |
+      | Money lives: Paul                                 |
+    And I should be on page "1" of "36" of the search results
+    And I should not see the previous button
+
+  Scenario: Search and navigate to the next result page
+    When I search the query "money"
+    And I go to the next results page
+    Then I should see "10" results
+    And I should see the search results
+      | title                          |
+      | Debt advice evaluation toolkit |
+    And I should be on page "2" of "36" of the search results
+
+  Scenario: Search and navigate to last result page
+    When I search the query "money"
+    And I go to the last page of results
+    Then I should be on page "36" of "36" of the search results
+    And I should see the search results
+      | title                                                        |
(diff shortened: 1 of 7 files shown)


features/search.feature
```diff
@@ -1,13 +1,36 @@
...
+
+  Scenario: Search and navigate to the next result page
+    When I search the query "money"
+    And I go to the next results page
+    Then I should see "10" results
+    And I should be on page "2" of "36" of the search results
+
+  Scenario: Search and navigate to last result page
+    When I search the query "money"
+    And I go to results page "35" of the "money" search query
```
willhall88: I'd still wonder whether to reword this step. I think it could confuse the business logic of the feature to take into account zero indexing. It isn't currently a reusable step by the look of it and so could easily be reworded. wdyt?

features/search.feature
```diff
@@ -1,13 +1,36 @@
...
-    Then I should see the search box
-
-  Scenario: Search can be performed
+  Background:
     Given I am on the home page
-    When I search for something relevant
-    Then I should see search results
+
+  Scenario: Search on first result page
+    When I search the query "money"
```
Tr4pSt3R: There needs to be a data setup process here. 

```gherkin
Given I am on the home page
And I have ten documents related to money
When I search for the term "money"
```

Running the tests without the data setup phase is jump which you may want to rethink?

features/search.feature
tomas-stefano: Thank you for the clarification. But the "35" would be hardcoded on the step. Then it could trick who uses the step *in case the last page is not 35*. Am I missing something?

willhall88: ok, I see your point. I think because it currently isn't being reused it would be ok to be hardcoded. If we want to reuse it in the future we can refactor? we could always have a comment as to why it is 35 as well?

tomas-stefano: ♻️

tomas-stefano:
@margOnline yes you are right. it is worth checking that the search term is included on the page results. Although I don't think multi word search term should be tested because to me this is not an extra scenario of the frontend app. Saying that I addressed all comments ♻️

margOnline:
Do you think its worth checking that the search term is on some of the other pages? maybe just the last one? 

Re multi word searches - I'm not sure what you mean by `an extra scenario of the frontend app'. If its a scenario that needs testing, isn't here the appropriate place to add it since tech debt is being addressed?

tomas-stefano:
Yes it makes sense we test the results for other pages.

About the multi wording my point is: it is not another scenario to test. IMHO it would be best to don't test any Algolia behaviour. Also, yesterday I had a meeting with people in the Digital team and they still need to decide how to approach and configure the Algolia behaviour, so if we test any specific Algolia behaviour could lead to unstable tests until the business decides how the search will behave.

PR #1935, opened 2018-04-13
[DEV APPROVED] - Replace rubocop with mas-standards rubocop

tomas-stefano:
**Target Process ticket**

No ticket. Done on development learn time.

**Summary**

As I like challenges I decided to add our rubocop from [mas-standards](github.com/moneyadviceservice/mas-standards) repo into the frontend repo.

There was 1062 offences with the old rubocop and with the new a litle bit more. I fixed almost all of them but there are still almost 66 offences.

Another thing that I realised is that a lot of code not being used by the frontend. I really really like to delete them but this should be another work/PR on its own. Most of unused files lived under **lib/core** dir so I Ignored them on rubocop for now.

Also I decided to continue to the LineLength as 120 for now and I will return to 80 and fixed the offences in another PR as well in a next DLT.

**QA / Testing**

There should be some QA manual tests which I will do and ask a QA to do it as well.

**Checklist**

+ [x] Tests passed.

### DIFF
--- a/app/controllers/concerns/chat.rb
+++ b/app/controllers/concerns/chat.rb
@@ -3,8 +3,8 @@ module Chat
 
   included do
     helper_method def chat_opening_hours
-                    @chat_opening_hours ||=
-                      ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
-                  end
+      @chat_opening_hours ||=
+        ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
+    end
   end
 end
--- a/app/controllers/content_hub_controller.rb
+++ b/app/controllers/content_hub_controller.rb
@@ -1,14 +1,12 @@
 class ContentHubController < ApplicationController
-
   layout 'content_hub'
 
-  def show
-  end
+  def show; end
 
   private
 
   def translation_prefix
-    "content_hub.#{params[:slug].gsub(/-/, '_')}"
+    "content_hub.#{params[:slug].tr('-', '_')}"
   end
 
   helper_method :translation_prefix
--- a/app/controllers/cost_calculator_builder_controller.rb
+++ b/app/controllers/cost_calculator_builder_controller.rb
@@ -6,7 +6,7 @@ class CostCalculatorBuilderController < MountController
   end
 
   # Note: syndicated_tool_request? is overridden in the engine.
-  #       This tool isn't served on the partner-tools subdomain,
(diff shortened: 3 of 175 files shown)


app/controllers/concerns/chat.rb
```diff
@@ -3,8 +3,8 @@ module Chat
 
   included do
     helper_method def chat_opening_hours
-                    @chat_opening_hours ||=
-                      ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
+      @chat_opening_hours ||=
+        ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
                   end
```
willhall88: this is an awkward indentation! I can see why the above lines were indented so much. if you move this `end` left as well, will it complain?

app/controllers/content_hub_controller.rb
```diff
@@ -1,14 +1,12 @@
...
 
-  def show
-  end
+  def show; end
 
   private
 
   def translation_prefix
-    "content_hub.#{params[:slug].gsub(/-/, '_')}"
+    "content_hub.#{params[:slug].tr('-', '_')}"
```
willhall88: I learnt something new with this one!

app/controllers/cost_calculator_builder_controller.rb
```diff
@@ -6,7 +6,7 @@ def category_id
   end
 
   # Note: syndicated_tool_request? is overridden in the engine.
-  #       This tool isn't served on the partner-tools subdomain,
+  #        This tool isn't served on the partner-tools subdomain,
```
willhall88: ❓ 🤔

app/helpers/budget_warning.rb
```diff
@@ -14,7 +14,7 @@ def initialize
     end
 
     def go_live?
-      Time.zone.today >= date ? true : false
```
willhall88: 🤦‍♂️

tomas-stefano: haha same reaction from me.

app/controllers/concerns/chat.rb
tomas-stefano: It was complaning actually:
```
app/controllers/concerns/chat.rb:8:19: W: Layout/DefEndAlignment: end at 8, 18 is not aligned with helper_method def at 5, 4.
                  end
```
I put the end on the right place and now we have 65 offences :D.

PR #1959, opened 2018-07-17
Fix reset password token

tomas-stefano:
[TP card](https://moneyadviceservice.tpondemand.com/entity/9322-reset-password-token-is-invalid)

## Context

When a user follows the journey to reset their password, at the point where a new password is entered and submitted, the following error is displayed and the task can not be completed - 'Reset password token is invalid'.
 
After implementing the encryption PR #1948, the reset password token stopped working.

The reason is the overwrite of the `find_first_by_auth_conditions`
 method on the user model. This method is used  by **Devise** for authentication and **_also for
the reset password token._** The original change to the method resulted in a search **only  by email**, hence searching by token was no longer working.

Making the `conditions` more general makes both scenarios work.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -57,7 +57,7 @@ RSpec.describe User, type: :model do
 
       context 'when token does not exist' do
         let(:conditions) do
-          { reset_password_token: "#{user.reset_password_token}C" }
+          { reset_password_token: "invalid_#{user.reset_password_token}" }
         end
 
         it 'returns nil' do
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -16,6 +16,57 @@ RSpec.describe User, type: :model do
     end
   end
 
+  describe '.find_first_by_auth_conditions' do
+    let!(:user) { create(:user, reset_password_token: SecureRandom.uuid) }
+    subject(:find_first_by_auth_conditions) do
+      User.find_first_by_auth_conditions(conditions)
+    end
+
+    context 'when searching by email' do
+      context 'when email exists' do
+        let(:conditions) do
+          { 'email' => user.email }
+        end
+
+        it 'returns user with same email' do
+          expect(find_first_by_auth_conditions).to eq(user)
+        end
+      end
+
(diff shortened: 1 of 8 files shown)


spec/models/user_spec.rb
```diff
@@ -16,6 +16,57 @@
...
+        end
+
+        it 'returns user with same token' do
+          expect(find_first_by_auth_conditions).to eq(user)
+        end
+      end
+
+      context 'when token does not exist' do
+        let(:conditions) do
+          { reset_password_token: "#{user.reset_password_token}C" }
```
willhall88: is the `C` supposed to be in the string here?

willhall88: or is this to make it an invalid token?

tomas-stefano: Yes I did for invalid token but I am happy to put anything else. Wdyt?

margOnline: In the interests of clarity, it would probably be better to make that string more representative of its purpose.

tomas-stefano: I liked the "invalid_" solution. I already pushed ♻️

tomas-stefano:
@margOnline I think the documentation is a done criteria for the TP card but not for this PR since the users are being affected by. Since I didn't work on the encryption neither the devise method I feel that I am not the right person to document this.  Perhaps @willhall88 ?

margOnline:
@tomas-stefano I'm not sure I agree with the comment re 'done criteria' but equally, I think we need to get this deployed asap. I would hate for the documentation requirement to not get done, perhaps a card can be raised and given a priority?

margOnline:
@tomas-stefano what happened to the cucumber test?

tomas-stefano:
What exactly you don't agree? I agreed that a documentation should be done but it is not me that should do it, that's all. I just jump in yesterday to fix a bug on a work that I was not involved at all and don't understand completely and that's the reason I think I should not be doing the documentation of the encryption and Devise. About your commit, I pushed with force without the lease option. Rookie mistake to someone that wants to finish this work asap to return to fincap. Do you mind to push again, please? I really sorry.

PR #1964, opened 2018-07-27
[DEV APPROVED] Add pacs engine to frontend

margOnline:
This PR adds the pacs engine latest version (0.35.0) to the MAS site **behind a http authentication**.

**Checklist**

+ [x] Tests passed.


config/schedule.rb
```diff
@@ -0,0 +1,3 @@
+every 1.day, at: '4:30 am' do
+  rake 'feed:store' # from the PACs engine
```
your: @osharpe is invoking `rake` enough or do we need to have a custom command to cd into the right directory and so on?

osharpe: @your you will need to cd and source environment variables to make this work. Something like `cd /srv/frontend/ && source /etc/mas/environment && RAILS_ENV=production bundle exec rake feed:store` We will also need to make some puppet changes to setup the whenever gem cron on the hosts as currently its only setup on CMS host in Prod, this is something I can help with in due course.

your: @osharpe makes sense, I'll keep that in mind. I can amend at least the first part in the while, thanks :)

app/controllers/pacs_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def authentication_required?
+    Authenticable.staging? || Rails.env.production?
+  end
+
+  def category_id
+    nil
+  end
+  helper_method :category_id
+end
```
your: I know it's meant to be temporary the authentication bit, does it mean we are ok to deploy it without a test to validate it? 😞

config/mongoid.yml
```diff
@@ -0,0 +1,153 @@
...
+    # use_utc: false
+
+    # Set the Mongoid and Ruby driver log levels when not in a Rails
+    # environment. The Mongoid logger will be set to the Rails logger
+    # otherwise.(default: :info)
+    # log_level: :info
+
+    # Application name that is printed to the mongodb logs upon establishing a
+    # connection in server versions >= 3.4. Note that the name cannot exceed 128 bytes.
+    # app_name: MyApplicationName
```
your: I was thinking maybe we can clean this a bit and remove all the comments? It makes it hard to read otherwise...

config/schedule.rb
```diff
@@ -0,0 +1,3 @@
+every 1.day, at: '4:30 am' do
+  command "cd /srv/frontend/ && source /etc/mas/environment && RAILS_ENV=production bundle exec rake feed:store"
```
your: The `job_type` didn't work?

tomas-stefano: @osharpe mentioned his preferred approach to have the same as cms (which confirmed works). Since I didn't had strong opinions I just follow the approach.

app/controllers/pacs_controller.rb
your: I get that @tomas-stefano, I personally think that testing is just part of our workflow, no matter if the code is temporary (unfortunately), even if I've raised the question mainly because we don't want to expose this by accident before launch. But if as you said you tested manually and feel comfortable with it, that's fair enough to me. 👍

tomas-stefano: Yes I agree with you, tests should be part of our workflow. I don't want to make a rule though about temporary things to not have tests because it will depend on each case. I think for the nature of this "Have a temporary http authentication until the data is confirmed by the banks that is accurate" it sounds very heavy to me that I shouldn't add tests for this. 
Other case to mention, If the business came with "temporary" things then I would not trust as much of the temporary and I would probably add tests, but PACS needs to be public at some point which makes me feel that's the reasonable approach.

margOnline: I'm not sure I agree with not having tests for this feature and the reasons given do not convince me otherwise; plus it's a dangerous precedent to set. I would have preferred to have a discussion or a bit more time to think it through. I also think that when going against the agreed code practices of the team, it would be better to discuss with the whole team. For one thing, it would be a very good way of stating that this is not setting a precedent.

scruti: Probably I shouldn't have approved it with an ongoing discussion :bowing_man:

tomas-stefano: I understand everyone, my intentions was not to set a precedent as well. I put this on the commit messages believing that everybody will read but I think I failed on that assumption. I will move to PR description before.
But I really like to heard what are the flaws of my arguments (but be nice please!). Anyway, if everyone still wants a test I am happy to add a PR adding one, even if I disagree.

config/schedule.rb
osharpe: ye wasn't too sure about the carriage return, so thought prob best to go with what we know works. We can of course test this `job_type` in test environment and use it if it add some value.

app/controllers/pacs_controller.rb
tomas-stefano: By the way, sorry If I merged this PR. The time that I merged the PR that was two comments only and Giuseppe approved so then I merged the PR, but I am happy to talk about this.

PR #1981, opened 2018-10-10
Updates JS to avoid URL Web API, which is not compatible with IE

davidtrussler:
[TP9686](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=bug/9686)

Partner tools currently do not display on IE. 

This appears to be because the [URL Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Browser_compatibility) used is not compatible with IE (all versions). 

This fix updates the JS to avoid using this. 

Testing is not easy since this fix needs to be deployed first but it can be done when this is on a Test environment by running a local page along the lines suggested in the comments in the TP ticket.


osharpe:
@davidtrussler Did you try deploying this to a test environment? We did some updates to `tools.js` recently to make partner-tools work across the test environments

davidtrussler:
@osharpe No I haven't done so but it would make sense. I'll do that now. Cheers.

osharpe:
Probably best to use UAT

davidtrussler:
Was about to (but only cos it was free). Good point though.

PR #1985, opened 2018-10-24
[DEV APPROVED] - 9703 PACs – introduce caching

your:
**Target Process ticket**
[TP 9703](https://moneyadviceservice.tpondemand.com/entity/9703-pacs-be-implement-caching)

**Summary**
Bumps the pacs engine to version `1.0.0` to include caching.

**QA / Testing**
In testing environments the cache generation has to be triggered manually by a backend dev.

Related: https://github.com/moneyadviceservice/pacs/pull/97

Depends on: https://github.com/moneyadviceservice/scripts/pull/1334


.env-example
```diff
@@ -30,3 +30,4 @@ DEFAQTO_API_TIMEOUT=
 MONGO_HOSTNAME=
 PACS_MONGOD_HOST='http://localhost:27017/'
 PACS_MONGOD_DB='frontend_development'
+PACS_CACHE_ENABLED=false
```
tomas-stefano: I am pretty sure is on your radar but just in case ... don't forget to add the var on scripts repo. ;)

your: Yeah, it is indeed. Will have a PR later on, thanks for reminding it anyway. 🙇

PR #2006, opened 2018-11-20
Bump debt advice locator 2.9.1

osharpe:
Release Debt advice locator 2.9.1


tomas-stefano:
I will consider as Dev approved since we usually don't review bump versions (unless different people work on that version).

PR #2018, opened 2018-12-10
update urls for pdf downloads

margOnline:
**Target Process ticket**

[TP 9911](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiQzgxRjcwQTZCMjdGNDBDNzY0NzgxOEUwRjBFMjQ3NkQifQ==&boardPopup=userstory/9911/silent)

**Summary**
The download link to the printed guide for the employer best practice pdf is incorrect.

**Checklist**

+ [x] Tests passed.


MarcoBgn:
Regarding the `package-lock` I have found an interest contribution:
```

> the npm documentation and the CLI message should at least make it clearer that using a package-lock.json comes with a serious drawback: the entire tree of dependencies and sub-dependencies will not be updated anymore, therefore bug fixes and new feature won't be installed automatically.
> 
> I think this worth mentioning clearly to users as downloading automatically the last fixes and features, (thanks to npm embrace of semantic versionning) is one of the most important feature of npm and it is nullified by using a package-lock.json.
> 
> I also noted that a lot of users are confused about why they experience a bug due to a sub-dependency while it has already been fixed and released.
> For example if the project some-project uses package-main which it self uses package-dep as its dependency and package-dep has a bug affecting package-main and therefore some-project.
> When package-dep releases a fix (with a patch release), if some-project has a package-lock.json it's owner will never get the fix. It's quite common that in such case the owner of some-project would open an issue on the package-main to ask to explicitly to upgrade to a version of package-dep with the fix and to make a new release of package-main so finally some-project can be updated to use it.
> See semantic-release/apm-config#35 for an example.

```

Full discussion here: https://github.com/npm/npm/issues/20603

margOnline:
@MarcoBgn good spot, package-lock.json file was not supposed to get checked in 😞 I have removed it.

PR #2020, opened 2018-12-11
Rake task to update some corporate_partner records:

margOnline:
**Target Process ticket**
[TP 9641](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiQzgxRjcwQTZCMjdGNDBDNzY0NzgxOEUwRjBFMjQ3NkQifQ==&searchPopup=userstory/9641)

**Summary**
The embed code that is generated by the Syndication page for the Pension Calculator is incorrect in that it includes a link to `https://www.moneyadviceservice.org.uk/en/tools/pensions-calculator` which should be `https://www.moneyadviceservice.org.uk/en/tools/pension-calculator`. The error is due to an incorrect value on a database field.

**Solution**
One off rake task to correct the value of the tool_name on the CorporatePartner table:
- Find records where the tool name = 'Pensions calculator'
- Update the tool_name to 'Pension calculator' and save the record

 Once run, this task should be deleted.

**Checklist**

+ [x] Tests passed.


lib/tasks/update-pension-calc-tool-name.rake
```diff
@@ -0,0 +1,18 @@
+namespace :corporate_partners do
+  desc 'Update tool name from "Pensions calculator" to "Pension calculator"'
+  task update_tool_name: :environment do
+    records_to_update = CorporatePartner.all.select do |partner|
+      partner.tool_name == 'Pensions calculator'
+    end
+
+    CorporatePartner.transaction do
+      records_to_update.each do |record|
```
scruti: you can avoid doing 5x DB updates with a `records_to_update.update_all(tool_name: 'Pension Calculator')`, that will execute a single DB update.

margOnline:
@scruti good suggestions, I need to do more db work as I'm forgetting all the activerecord methods, ♻️

scruti:
re-approved!

lib/tasks/update-pension-calc-tool-name.rake
```diff
@@ -0,0 +1,13 @@
+namespace :corporate_partners do
+  desc 'Update tool name from "Pensions calculator" to "Pension calculator"'
+  task update_tool_name: :environment do
+    records_to_update = CorporatePartner.where(tool_name: 'Pensions calculator')
+    puts "record ids: #{records_to_update.map(&:id)}"
+
+    CorporatePartner.transaction do
+      records_to_update.update_all(tool_name: 'Pension calculator')
```
your: Note: `update_all` skips all `ActiveRecord` validations and does not update the `updated_at` field, which can be very handy in situations like this actually. On the top of that, it produces an atomic SQL statement, that means it does not need to be wrapped in a transaction block. It will also return the number of records updated.

PR #2028, opened 2019-01-04
Reorganise MAS-maintained gems in Gemfile

malcolmbaig:
Among the ungrouped gems, list the MAS ones separately. This is useful
for identifying which repos need updating when applying a change across
all Frontend dependencies, eg - a Ruby version bump.


malcolmbaig:
@margOnline Yes, I think it is. Added a commit.

PR #2058, opened 2019-02-06
[DEV APPROVED] 10212 MAS pinterest verification tag

rhianwatkins1:
[TP10212](https://moneyadviceservice.tpondemand.com/entity/10212-get-pinterest-verified)

Really straightforward bit of work just to add in a meta tag generated by pinterest so that they can verify MAS

### DIFF
--- a/app/views/layouts/_base.html.erb
+++ b/app/views/layouts/_base.html.erb
@@ -25,6 +25,7 @@
   <meta name="HandheldFriendly" content="True">
   <meta name="application-name" content="Money Advice Service">
   <meta name="msapplication-starturl" content="https://www.moneyadviceservice.org.uk">
+  <meta name="p:domain_verify" content="96b592752f8004b8838349b27c7355ba"/>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_base.html.erb
```diff
@@ -186,7 +186,7 @@
 
   </script>
   <%# END MAS JAVASCRIPT SETUP %>
-
+  <meta name="p:domain_verify" content="96b592752f8004b8838349b27c7355ba"/>
```
willhall88: does this have to go at the bottom of the `<head>` or can we put it at the top with all the other meta tags?

davidtrussler: I think I would prefer to see this with the other tags as well. I don't believe there would be any effect on performance. Not critical though since this addition is temporary.

rhianwatkins1: @willhall88 @davidtrussler 👍 moved

PR #2064, opened 2019-02-15
[DEV APPROVED] 10107 Upgrade to Ruby 2.5.3

malcolmbaig:
[TP](https://moneyadviceservice.tpondemand.com/entity/10107-upgrade-frontend-repo-to-ruby-253)


Gemfile
```diff
@@ -8,16 +8,21 @@ source 'http://gems.dev.mas.local'
...
 
+ruby '2.5.3'
 gem 'rails', '4.2.11'
 
 gem 'activerecord-session_store'
-gem 'aes'
+# The aes gem is no longer supported, so point to a fork that resolves
+# deprecation warnings in Ruby's OpenSSL module. This could feasibly be
+# replaced by an implementation within this repo, as all the gem is doing is
+# wrapping OpenSSL.
```
scruti: nice explanation

Gemfile.lock
```diff
@@ -901,5 +911,8 @@ DEPENDENCIES
   whenever
   wpcc (= 2.2.0)
 
+RUBY VERSION
+   ruby 2.5.3p105
```
scruti: is the PATCH related with the issue when running it on MAC OSX?

malcolmbaig: No, I don't think so - that's just the expected patch level for the 2.5.3 release. Doesn't have any bearing on the segfault issue, AFAIK.

malcolmbaig: @scruti your question got me wondering what the point of these patch levels is, given that Ruby switched to a form of semantic version a while ago. Found this, which I think is still accurate (see 'Version schema'): https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/

PR #2066, opened 2019-02-20
[DEV APPROVED] Remove dry-configurable version lock

scruti:
In order to merge Pacs gem upgrade, we had to lock the version of
dry-configurable dependency gem to 0.7.x, as the one coming with
Pacs version bump (dry-configurable 0.8.1) was dependant on ruby >= 2.3

Now that Frontend is using Ruby 2.5.3, this lock can be removed and
we can bump dry-xxx gems


Gemfile
```diff
@@ -25,9 +25,6 @@ gem 'devise', '~> 4.4.0'
 gem 'devise-encryptable'
 gem 'dotenv'
 gem 'draper', '< 3'
-# TODO: Remove this dry-config line all together once bump to ruby v2.5 as only is
-# needed to lock it due to clashing between newer versions and ruby < v2.3
-gem 'dry-configurable', '~> 0.7.0'
```
margOnline: 🙌

PR #2074, opened 2019-03-07
[WIP] 9754 update jquery 2 part nbm

nmiddleweek:
For review.

I'll clear out the rogue commits tomorrow.

### DIFF
--- a/app/assets/javascripts/components/JQueryMigrateWrapper.js
+++ b/app/assets/javascripts/components/JQueryMigrateWrapper.js
@@ -1,0 +1,33 @@
+define(['jquerySrc', 'jquerymigrate'],
+  function(jQuery, jquerymigrate) {
+    'use strict';
+
+    /**
+     * Module Name: JQueryMigrateWrapper
+     *
+     * Description: This merges jQuery Migrate v3 with jQuery 331 and provides a jQuery 2 backwards compatible version.
+     * Using this to override the 'jquery' entry in the require_config.js.erb means you no longer
+     * have to update every module to load the jQueryMigrate in separately.
+     *
+     * In the require_config, we following these steps:
+     *
+     * 1. Change the jquery path to jquerySrc - this temporarily invalidates all references to 'jquery'.
+     * 2. we load the jquerymigrate: requirejs_path('jquery-migrate/jquery-migrate') (as we have been).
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/components/JQueryMigrateWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+     * 4.   jquerymigrate: ['jquerySrc'],
+     * 5.   jquery: ['jquerymigrate'],
+     *      .
+     *      .
+     *      .
+     *    }
+     *
+     */
+
+    console.log('jQuery.migrateVersion', jQuery.migrateVersion, 'jQuery.fn.jquery', jQuery.fn.jquery);
```
davidtrussler: I can't see this being logged anywhere. Is it useful?

davidtrussler:
@nmiddleweek I'm not sure why this is in a separate branch. I think it would be clearer to make the jQueryMigrateWrapper commit to the existing one. Also a fuller description would be useful to explain what this is doing.

app/assets/javascripts/components/JQueryMigrateWrapper.js
```diff
@@ -0,0 +1,33 @@
...
+     *
+     *    shim: {
+     * 4.   jquerymigrate: ['jquerySrc'],
+     * 5.   jquery: ['jquerymigrate'],
+     *      .
+     *      .
+     *      .
+     *    }
+     *
+     */
```
davidtrussler: Can we add a comment that this should be deleted when the migrate work is done, and also a comment in require_config to document that that should be updated also at the end of the process?

willhall88:
@davidtrussler So I asked Nick to upload the new commit to a new branch, but I think it was lost in translation. I just intending on it being a single commit on it's own branch as I was going to then take that forward with the progression through 1.11 -> 1.12 -> 3 that we need to go through and cherry pick over some of the other commits as needed. Rather adding these changes (like rolling back to 1.12 etc.) to the existing branch

willhall88:
We also found that adding the setup with the wrapper module should trigger the jqmigrate on all pages on frontend, which we weren't seeing on non tool pages. I will update in the channel too

willhall88:
closing this in favour of #2076

PR #2087, opened 2019-03-19
[DEV APPROVED] Add pinterest verification file

margOnline:
[TP 10294](https://moneyadviceservice.tpondemand.com/restui/board.aspx#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiRkRBMzk3RTBGNjU3NDg1RDM4QzczMTJFNzI5MDFBN0MifQ==&boardPopup=userstory/10294/silent)

Add a file to enable pinterest verification via email.


margOnline:
> Looks good. Just curious tho, why pacs was updated in this?

@rhianwatkins1 I mistakenly added the pinterest file to pacs and upped that engine version to bring in the file, so to remove the file from pacs I need to up the version again 🤦‍♀️  I could just revert back to the previous pacs version but its better practice to do it this way.

PR #2099, opened 2019-04-01
[DEV APPROVED] 10403 Update footer reference from SFGB to MPS

rhianwatkins1:
**Target Process ticket**
[TP10403](https://moneyadviceservice.tpondemand.com/entity/10403-update-mas-footer-reference-from-sfgb)

**Summary**
This PR swaps over the sfgb logo in the footer to be the new Money & Pensions Service logo.
- Updated logo image
- Updated text & .yml files
- Updated the link
- Renamed the sfgb_logo sass file

Please note, that it was required for the text to be on 3 lines. To do this, 3 sepereate lines were created in the translation files. Because browsers render font differently, I couldn't do this by setting the width. Also due to English and Welsh being different lengths, Welsh had to break differently, so the first and second lines had to be split. This means the translation file looks a bit messy, but we know it will render correctly for all users.
Also, I've taken more of a "bilingual first" approach, therefore the images are of the same size, with the english one containing more white space, this is to avoid things like the logo getting cropped in Welsh as happened with the MAS logo. 

I've not changed any of the responsive styles, breakpoints are as were previously.

**Screenshots**
Before EN:
Screenshot 2019-04-01 at 12 37 25
Before CY:
Screenshot 2019-04-01 at 12 39 16
After EN:
Screenshot 2019-04-01 at 12 43 43
After CY:
Screenshot 2019-04-01 at 12 43 58

**Checklist**

+ [ ] Tests passed.
+ [ ] Rubocop is passing for this PR (or any other lint like JShint).
+ [ ] Code Climate passed for this PR.


rhianwatkins1:
Tests are failing, will look tomorrow as it's coming from a tool unrelated to my changes.

MarcoBgn:
@rhianwatkins1 :

> Tests are failing, will look tomorrow as it's coming from a tool unrelated to my changes.

This is more than likely due to the local ruby installation. Everything is fine on my machine

app/assets/stylesheets/components/common/_footer_secondary.scss
```diff
@@ -104,14 +104,20 @@
...
 }
 
-.footer-secondary__sfgb-text {
+.footer-secondary__mps-text {
   display: inline-block;
+}
+.footer-secondary__mps-text--en {
+  width: 98px;
   & > * {
     display: block;
```
scruti: :question:  this `display: block` declaration being only present in the `EN` version is something we need to have or we forgot to copy it? (or may not be necessary at all).

app/assets/stylesheets/components/common/_footer_secondary.scss
```diff
@@ -104,14 +104,20 @@
   }
 }
 
-.footer-secondary__sfgb-text {
+.footer-secondary__mps-text {
   display: inline-block;
+}
+.footer-secondary__mps-text--en {
+  width: 98px;
```
davidtrussler: I'm not sure about giving this an explicit width because it relies too much on the text displaying pretty much the same across browsers and devices. Would it be better to follow the current markup that divides the text into three sections? (P.S. I don't like that either so there doesn't seem to be a prefect solution).

app/views/shared/_footer_secondary.html.erb
```diff
@@ -28,14 +28,13 @@
...
-          <a class="footer-secondary__link" href="https://singlefinancialguidancebody.org.uk/" target="_blank">
-            <span class="footer-secondary__sfgb-text">
-              <span><%= t('footer.sfgb_part_of') %></span>
-              <span><strong><%= t('footer.sfgb_single_financial') %></strong></span>
-              <span><strong><%= t('footer.sfgb_guidance_body') %></strong></span>
+        <li class="footer-secondary__list-item footer-secondary__mps-link">
+          <a class="footer-secondary__link" href="https://http://moneyandpensionsservice.org.uk/" target="_blank">
+            <span class="footer-secondary__mps-text footer-secondary__mps-text--<%= locale %>">
+              <span><%= t('footer.mas') %></span>
+              <span><%= t('footer.mps_provided') %></span>
```
davidtrussler: See comment above about splitting the text.

davidtrussler:
I left a couple of comments about how the text is being split.

Apart from that I think this does what it needs to from the ticket but I sill have a couple of queries on it about stuff that may have been introduced in the original logo change (though I appreciate these may be outside the scope of this piece of work): 
- I think the "Contact us, Terms & conditions, Privacy notice ..." links should display a full width on smaller viewports. 
- On larger viewports I don't think those links and the Accessibility link should be wider than the standard width of the rest of the page.

davidtrussler:
BTW I also get test errors on frontend but those are pre-existing and, as suggested by others, probably not caused by this work.

app/assets/stylesheets/components/common/_footer_secondary.scss
rhianwatkins1: So the reason I've done it this way is because it has to break differently on the welsh to the english because they're different lengths. Lets chat in person about this, I'm also not sure about it

app/assets/stylesheets/components/common/_footer_secondary.scss
rhianwatkins1: Nope, that's because the text in English and Welsh is different lengths and needs to break differently. I'm going to chat with @davidtrussler about this today and get his opinion, this may well change, will update here when we come to a conclusion

rhianwatkins1:
@davidtrussler I totally agree with you and I've left a little comment on the ticket, basically we're doing the bare minimum due to a deadline, and those changes are no longer in the scope of this ticket.

app/assets/stylesheets/components/common/_footer_secondary.scss
rhianwatkins1: @scruti So I've chatted to @davidtrussler about this, and we think the best way to do this for the user is to set the line breaks in the translation files. I've committed with those changes, which has removed this code entirely

app/assets/stylesheets/components/common/_footer_secondary.scss
rhianwatkins1: @davidtrussler I've committed the changes we discussed. Will update the PR description now to explain

PR #2113, opened 2019-05-13
[DEV APPROVED] 10484 Implementation of Floating Chat Button

andreferreiradlw:
**Target Process ticket**

[TP 10474](https://moneyadviceservice.tpondemand.com/RestUI/Board.aspx#page=board/5253959470622112123&appConfig=eyJhY2lkIjoiNTM0ODcyOThDQjBDMTMyMDgyMTBENEZEM0QwM0ZBQzkifQ==&boardPopup=userstory/10484/silent)

Raised button in article pages logic in Dough [PR 328](https://github.com/moneyadviceservice/dough/pull/328)

**Summary**
Added a floating button on the bottom right corner of the screen for mobile users only using the SVG chat icon.

**QA / Testing**
Testing for the raised button in article pages should refer the PR in Dough.


app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -19,25 +14,61 @@
...
+  
+  .chat-box__icon {
+    margin-left: -35px;	
+  }
 }
 
-.chat-box__icon {
-  margin-left: -35px;	
+.chat-popup__container {
+  bottom: 2%;
```
davidtrussler: Could we use a multiple of $baseline-unit for this value?

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -19,25 +14,61 @@
...
+  right: 2%;
```
davidtrussler: Can we use the $default-gutter variable for this value?

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: Corrected

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: Corrected

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -1,13 +1,8 @@
-.chat-box__container {
+.chat-box__container,
+.chat-popup__container {
   background-color: $color-white;
   position: fixed;
-  top:30%;
-  left: 0;
-  padding: $baseline-unit*3;
   z-index:9999;
```
msleiman: z-index is too high please consider a value under 100

app/views/shared/_contact_panels.html.erb
```diff
@@ -119,6 +119,12 @@
 </a>
 <% end %>
 
+<a href="#" aria-hidden="true" id="js-chat-popup-mobile" class="chat-popup__container" data-dough-component="ChatPopup">
+    <svg>
+      <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#svg-icon--web-chat"></use>
+    </svg>
```
msleiman: Please add a fallback image.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: These styles are outside the scope of this ticket as it refers to the desktop chat button but I can make these changes anyway

margOnline: That's a fair comment @andreferreiradlw but we try to take the approach that if we see a candidate for refactoring whilst working on a file or repo, we make the change. If its a big refactor, you can choose whether to tackle the refactor as part of the ticket or create a new ticket. We do this because some of the code base is quite old and this approach helps to keep it current.

andreferreiradlw: @margOnline I totally agree I just missed this one as these styles were already there but I've changed it to 99 now.

app/views/shared/_contact_panels.html.erb
andreferreiradlw: Good spot, I've implemented a fallback now

app/views/shared/_contact_panels.html.erb
```diff
@@ -119,6 +119,13 @@
 </a>
 <% end %>
 
+<a href="#" aria-hidden="true" id="js-chat-popup-mobile" class="chat-popup__container" data-dough-component="ChatPopup">
```
davidtrussler: It is more in keeping with our standards to avoid using id here but instead to use a data attribute. So in this case we could use data-dough-chat-popup-mobile or similar.

andreferreiradlw: I've removed the id property from this element as it's not needed anymore.

PR #2122, opened 2019-05-29
[DEV APPROVED] 10485 popup channel select

andreferreiradlw:
[TP 10485](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=board/5253959470622112123&appConfig=eyJhY2lkIjoiNTM0ODcyOThDQjBDMTMyMDgyMTBENEZEM0QwM0ZBQzkifQ==&boardPopup=userstory/10485/silent)

This PR implements the opening and closing functionality to the webchat popup as well as the service select and buttons for WhatsApp and Webchat.

[PR 330](https://github.com/moneyadviceservice/dough/pull/330) on dough implements all the logic      and tests for this popup.


app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -1,11 +1,11 @@
 .chat-box__container,
-.chat-popup__container {
+.mobile-webchat__container {
   background-color: $color-white;
   position: fixed;
-  z-index:99;
-  -webkit-box-shadow:  0 0 7px 1px rgba(133,133,133,0.71);
-  -moz-box-shadow:  0 0 7px 1px rgba(133,133,133,0.71);
-  box-shadow:  0 0 7px 1px rgba(133,133,133,0.71);
+  z-index: 90;
```
margOnline: why has this been decreased?

app/decorators/chat_opening_hours_decorator.rb
```diff
@@ -31,6 +31,10 @@ def open?
     object.now_open?
   end
 
+  def closed?
+    !open?
+  end
+
```
margOnline: ❤️

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
-    }
+  if (typeof sWOTrackPage == 'function') {
+    sWOTrackPage();
 
+  <% #Track page enagagement(a visit lasting more than 30 seconds) %>
+      setTimeout("sWOInvite='';sWOResponse='N';sWOPage=(sWOPage || '')+'%3Ftimer%3Dtrue';sWOTrackPage();", 30000);
+  }
 </script>
 <% end %>
 <% end %>
```
margOnline: This file is a candidate for refactoring. For example, what is the difference between `if chat_opening_hours.open?` and `if chat_opening_hours.status_as_string == 'open'`? Are there opportunities to partialize any of the content?

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: I have lowered this value on the container because it was conflicting with the X button that shows when the popup is opened

app/views/shared/_contact_panels.html.erb
andreferreiradlw: @margOnline just noticed that all of these instances of `chat_opening_hours.status_as_string == 'open'` are on the WhatsApp contact panel which is going to be reworked on [**TP 10511**](https://moneyadviceservice.tpondemand.com/RestUI/Board.aspx#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiNTM0ODcyOThDQjBDMTMyMDgyMTBENEZEM0QwM0ZBQzkifQ==&boardPopup=userstory/10511/silent) so we can address these issues on that ticket and make the content consistent as well.

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
+        <option value="homes_mortgages"><%= t('contact_panels.chat_popup.dropdown_options.homes_mortgages') %></option>
+        <option value="budgeting_saving"><%= t('contact_panels.chat_popup.dropdown_options.budgeting_saving') %>
+        </option>
+        <option value="work_benefits"><%= t('contact_panels.chat_popup.dropdown_options.work_benefits') %></option>
+        <option value="pensions_retirement"><%= t('contact_panels.chat_popup.dropdown_options.pensions_retirement') %>
+        </option>
+        <option value="family_care"><%= t('contact_panels.chat_popup.dropdown_options.family_care') %></option>
+        <option value="cars_travel"><%= t('contact_panels.chat_popup.dropdown_options.cars_travel') %></option>
+        <option value="insurance"><%= t('contact_panels.chat_popup.dropdown_options.insurance') %></option>
+      </select>
```
davidtrussler: For the ```<option>``` tags we should use the values for ```clumps``` and an each loop as [here](https://github.com/moneyadviceservice/frontend/blob/c51f7a7ba0050919cb3bb70a27e27945473dab93/app/views/shared/_global_nav.html.erb#L4). These are CMS controlled and could change in future.

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
+  <div class="mobile-webchat--icon">
     <span class="icon icon--web-chat"></span>
     <svg class="icon-svg--web-chat">
       <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#svg-icon--web-chat"></use>
     </svg>
-</a>
+  </div>
+  <div class="mobile-webchat__form">
+    <h4><%= t('contact_panels.chat_popup.title') %></h4>
+    <div class="mobile-webchat__form-select--container">
```
davidtrussler: I think we should use a slightly different naming convention here and below: since this element is a descendent of ```mobile-webchat__form``` it should probably be marked up as ```mobile-webchat__form__select-container``` or maybe more succinctly as ```mobile-webchat__select-container```.

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
+        </option>
+        <option value="work_benefits"><%= t('contact_panels.chat_popup.dropdown_options.work_benefits') %></option>
+        <option value="pensions_retirement"><%= t('contact_panels.chat_popup.dropdown_options.pensions_retirement') %>
+        </option>
+        <option value="family_care"><%= t('contact_panels.chat_popup.dropdown_options.family_care') %></option>
+        <option value="cars_travel"><%= t('contact_panels.chat_popup.dropdown_options.cars_travel') %></option>
+        <option value="insurance"><%= t('contact_panels.chat_popup.dropdown_options.insurance') %></option>
+      </select>
+    </div>
+    <div class="mobile-webchat__form-button--container">
```
davidtrussler: Similarly to above this should probably be marked up as ```mobile-webchat__form__button-container``` or ```mobile-webchat__button-container```. WDYT?

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
+      <div class="mobile-webchat__form-button--spacer">
+        <a class="mobile-webchat__form-button--whatsapp" href="https://wa.me/447701342744" target="_blank">
```
davidtrussler: I would prefer to see this using the class `button` here (and below) to give this our base styles which can then be modified where necessary by further more specific classes. Otherwise we are tending to drift away from our standard styles into bespoke ones for new elements.

app/views/shared/_contact_panels.html.erb
```diff
@@ -113,111 +114,173 @@
...
+            <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#svg-icon--whatsapp"></use>
+          </svg>
+          <div class="form-button__text">
+            <span class="form-button__text-heading"><%= t('contact_panels.chat_popup.button_text') %></span>
+            <span class="form-button__text-service">WhatsApp</span>
+          </div>
+        </a>
+      </div>
+      <div class="mobile-webchat__form-button--spacer">
+        <a class="mobile-webchat__form-button--webchat<% if chat_opening_hours.closed? %> disabled<% end %>" target="_blank">
```
davidtrussler: as above with regards to the class `button`.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -81,11 +89,257 @@
...
+// other contact methods link
+.mobile-webchat__contact {
+  margin-top: 10px;
+  text-align: right;
+
+  .mobile-webchat__contact--link {
+    color: $color-black;
+    text-decoration: none;
+    font-weight: 700;
+    font-size: 0.8rem;
```
davidtrussler: Can we retain our base styles for links as far as possible? In this instance I think we should at least keep the base colour and hover state (underline).

app/views/shared/_contact_panels.html.erb
davidtrussler: @margOnline All the JS in here is third party and I'd be reluctant to do much to it right now. It would be nice to pull it out as a discrete component and refactor it but I think that is something to be done in the future.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
davidtrussler: We need to look into the z-index value a bit more because right now this is in front of the menu and the overlay that greys out the main page content when the menu is open. We may need to change those values as well to make everything work well together.

app/views/shared/_contact_panels.html.erb
andreferreiradlw: Clumps are now in use, @davidtrussler please have a look and see if the implementation is fine

app/views/shared/_contact_panels.html.erb
andreferreiradlw: @davidtrussler  I agree, I've now changed some class names in the buttons and select element so they are simpler and shorter than the previous names

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: That's good to me, I've now removed both the `color` and `text-decoration` markup

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -81,11 +89,257 @@
...
+
+  // popup opened state
+  & > *:not(.mobile-webchat--icon) {
+    opacity: 0;
+    transition: opacity 0.4s linear 0.1s;
+  }
+
+  &.mobile-webchat--opened {
+    padding: $baseline-unit*2;
+    width: 92vw;
```
davidtrussler: I think the width would be better expressed as a percentage so that it is more evenly positioned. In this case it could be something like `100% - ($default-gutter * 4)`.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -35,42 +36,49 @@
...
-  border-radius: 4px;
-  display: block;
-  transition: all 0.4s;
+  border-radius: .15em;
+  transition: transform 0.5s, bottom 0.4s, width 0.3s cubic-bezier(0.075, 0.82, 0.165, 1);
+  width: 76px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
davidtrussler: Good use of flexbox but I think we need to have a fallback for non-flexbox browsers (IE9 really). Nothing major, maybe just explicit widths. (N.B. I haven't checked what it renders like on such browsers).

app/views/shared/_contact_panels.html.erb
andreferreiradlw: I didn't realise we had a class of `button`, this is now in use on both the buttons

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: It does look more evenly positioned, thanks for the suggestion!

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: Even by just removing the flex all the elements look the same with no changes. I'm just using flex here to make sure the children are in a column direction but if the browser doesn't support flex then it'll still going to be looking the same.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -81,11 +89,246 @@
...
+  width: 30px;
+  height: 30px;
+  padding: 3px;
+  text-align: center;
+  color: $color-white;
+  border: none;
+  border-radius: 0 .15em 0 .15em;
+  background-color: $color-green-primary;
+  transition: color 0.25s, background-color 0.25s, opacity 0.4s linear 0.4s;
+  z-index: 91;
```
davidtrussler: This value will need looking at together with above - it is in front of the menu and grey overlay as it is.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
davidtrussler: OK 👍

app/views/shared/_contact_panels.html.erb
margOnline: @davidtrussler my comment was with reference to the markup and erb, not the js. I guess I should have positioned the comment at the beginning of the file 

@andreferreiradlw I kind of think that this would be an appropriate time to at least address the inconsistent erb code.

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: I spoke with John Player and the popup will need to show behind the menu and the grey overlay every time so I'll be looking to make this change in the next commit

app/assets/stylesheets/components/page_specific/_floating_chat.scss
andreferreiradlw: This issue relative to the menu and grey overlay has now been fixed

app/views/shared/_contact_panels.html.erb
andreferreiradlw: @margOnline I've corrected the inconsistencies now regarding those instances of `chat_opening_hours.status_as_string == 'open'`

PR #2144, opened 2019-08-02
[DEV APPROVED] 10600 update cookie consent message

davidtrussler:
[TP10600](https://moneyadviceservice.tpondemand.com/restui/board.aspx?#page=userstory/10600)

The work in this PR 
- updates the cookie consent message 
- replaces the close button with one that states that the user accepts the use of cookies on the site. 

Existing functionality remains unchanged. 

Current cookie message: 
image

Updated cookie message: 
image

### DIFF
--- a/app/assets/stylesheets/components/common/_cookie_message.scss
+++ b/app/assets/stylesheets/components/common/_cookie_message.scss
@@ -12,24 +12,19 @@
 
 .cookie-message__content {
   @extend %type-small;
-  @include column(11);
+  @include column(12);
   margin-top: $baseline-unit*3;
   margin-bottom: $baseline-unit*3;
 }
 
 .cookie-message__link {
   margin-top: $baseline-unit*2;
-  display: block;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/components/common/_cookie_message.scss
```diff
@@ -12,24 +12,29 @@
 
 .cookie-message__content {
   @extend %type-small;
-  @include column(11);
+  @include column(12);
   margin-top: $baseline-unit*3;
   margin-bottom: $baseline-unit*3;
+  padding-right: 76px;
```
andreferreiradlw: I believe this line should be deleted as your setting padding-right again in line 19

davidtrussler: Yeah, it's just there for browsers that don't understand `calc()`. Not critical perhaps, but would prefer to specify something in this case.

PR #2152, opened 2019-09-05
10695 popup pensions content

andreferreiradlw:
[TP10695](https://maps.tpondemand.com/entity/10695-replace-popup-chat-content-for-pensions)

This PR aims to replace the contents of the chat popup for pages which belong to the pensions category.  When a user opens the chat popup we will present them with specific webchat content.


app/views/shared/_webchat_popup.html.erb
```diff
@@ -9,18 +9,20 @@
     </svg>
   </div>
   <div class="mobile-webchat__form">
-    <h4><%= t('contact_panels.chat_popup.title') %></h4>
+    <% if pensions_and_retirement_page? %>
+      <%= render 'shared/webchat_popup_pensions' %>
+    <% else %>
+    <h4><b><%= t('contact_panels.chat_popup.title') %><b></h4>
```
davidtrussler: Do we need the `<b>` tag here? Doesn't seem to make much difference if it's removed. However, if we do, the end tag is wrong (`<b>` not `</b>`).

app/views/shared/_webchat_popup_pensions.html.erb
```diff
@@ -0,0 +1,30 @@
+<h4><b><%= t('contact_panels.chat_popup_pensions.title') %><b></h4>
+<h4><%= t('contact_panels.chat_popup_pensions.intro') %></h4>
```
davidtrussler: As above with the `<b>` tags (do we need them and the end tag is wrong). Not sure we should have 2 consecutive `<h4>`s either. Can we give these classes to deal with the styling if necessary?

app/views/shared/_webchat_popup_pensions.html.erb
```diff
@@ -0,0 +1,30 @@
...
+<div class="mobile-webchat__button-container">
+  <div class="mobile-webchat__button">
+    <a class="button mobile-webchat__button--pensions disabled">
```
davidtrussler: I assume this is disabled until the work in a further ticket.

davidtrussler:
@andreferreiradlw I've made a couple of minor comments here but otherwise I'm happy to say this is good to go. 👍

app/views/shared/_webchat_popup.html.erb
andreferreiradlw: I've removed this tag now and added in some styles for the bold title text.

app/views/shared/_webchat_popup_pensions.html.erb
andreferreiradlw: What should we have instead of the H4 tags?

PR #2154, opened 2019-09-11
10696 popup pensions display

andreferreiradlw:
**Target Process ticket**

[TP10695](https://maps.tpondemand.com/entity/10695-replace-popup-chat-content-for-pensions)

**Summary**
In ticket 10695 the contents of the chat popup were altered for pages in the pensions category, however, this popup only displays on mobile.  We would like to display the popup for desktop users who are viewing pensions content only, the chat popup will still be hidden on desktop for all other page types.

### DIFF
--- a/app/assets/stylesheets/components/page_specific/_floating_chat.scss
+++ b/app/assets/stylesheets/components/page_specific/_floating_chat.scss
@@ -1,19 +1,3 @@
-.chat-box__container,
-.mobile-webchat__container {
-  background-color: $color-white;
-  position: fixed;
-  z-index: 90;
-  box-shadow: 0 0 7px 1px rgba(133, 133, 133, 0.71);
-
-  &:hover {
-    text-decoration: none;
-  }
-
-  &:focus {
-    background-color: $color-white;
-  }
-}
-
 .chat-box__container {
   top: 30%;
   left: 0;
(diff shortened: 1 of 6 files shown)


app/views/shared/_webchat_popup.html.erb
```diff
@@ -1,4 +1,4 @@
-<div aria-hidden="true" class="mobile-webchat__container mobile-webchat--closed" data-dough-component="ChatPopup">
+<div aria-hidden="true" class="mobile-webchat__container mobile-webchat--closed <% if pensions_and_retirement_page? %>display-flex<% end %>" data-dough-component="ChatPopup">
```
davidtrussler: In fact the whole `mobile-webchat__container` class is rendered obsolete by this change. Maybe that just needs to be called `webchat__container` and removed from view on mobile except where the class is also `pensions` or some-such, which can replace the `display-flex` class. (See also above on the `display-flex` class).

app/assets/stylesheets/components/page_specific/_floating_chat.scss
```diff
@@ -72,6 +72,9 @@
   &:focus {
     outline: none;
   }
+  &.display-flex {
```
davidtrussler: I think we should use a more semantic class name than `display-flex`. (See also below on the `display-flex` class).

davidtrussler:
This looks good and works well. My only query is about naming now that this is no longer only just a mobile thing. I do realise though that if we started renaming stuff this might be quite a big task.

PR #2156, opened 2019-09-16
[DEV APPROVED] - Pensions Popup Opening Hours

andreferreiradlw:
[TP10697](https://maps.tpondemand.com/entity/10697-restrict-visibility-of-pensions-chat-popup)

This PR aims to restrict the visibility of the pensions popup in line with the TPAS opening hours.
9am to 6.20pm Monday to Friday

### DIFF
--- a/app/controllers/concerns/chat.rb
+++ b/app/controllers/concerns/chat.rb
@@ -6,6 +6,7 @@ module Chat
       @chat_opening_hours ||=
         ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
     end
+    
     helper_method def pensions_opening_hours
       @pensions_opening_hours ||=
         ChatOpeningHoursDecorator.decorate(Rails.application.config.pensions_opening_hours)
--- a/app/controllers/concerns/chat.rb
+++ b/app/controllers/concerns/chat.rb
@@ -6,5 +6,9 @@ module Chat
       @chat_opening_hours ||=
         ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
     end
+    helper_method def pensions_opening_hours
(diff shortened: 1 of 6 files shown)


app/views/shared/_contact_panels.html.erb
```diff
@@ -26,7 +26,7 @@
 </a>
 <% end %>
 
-<%= render 'shared/webchat_popup', locals: {clumps: clumps, chat_opening_hours: chat_opening_hours} %>
+<%= render 'shared/webchat_popup', locals: {clumps: clumps, chat_opening_hours: chat_opening_hours, pensions_opening_hours: pensions_opening_hours} %>
```
malcolmbaig: Is this supposed to be 'shared/webchat_popup_pensions', ie - the file below?

malcolmbaig: I think using locals here is a good thing for clarity, even though we don't need to because chat_opening_hours and pensions_opening_hours are helper methods that available globally in all view.

If we're using them here though, we should be consistent and pass them into the webchat_popup_pensions partial as well.

app/controllers/concerns/chat.rb
```diff
@@ -6,5 +6,9 @@ module Chat
       @chat_opening_hours ||=
         ChatOpeningHoursDecorator.decorate(Rails.application.config.chat_opening_hours)
     end
+    helper_method def pensions_opening_hours
```
malcolmbaig: Minor, but let's stick an empty line between these method definitions.

app/views/shared/_contact_panels.html.erb
andreferreiradlw: I wasn't too sure so I had removed the locals from the pensions partial. I've now put it back in, thanks!

PR #2173, opened 2019-12-11
[DEV APPROVED] 10975 maps banner

davidtrussler:
[TP10975](https://maps.tpondemand.com/entity/10975-add-maps-brand-banner-to-mas)

This work adds a MAPS banner to the MAS site. There are two related PRs for this: the one in Yeast, [PR57](https://github.com/moneyadviceservice/yeast/pull/57), adds the styles to the markup that is contained here. For testing both need to be loaded locally. 

The banner is rendered in three different sizes: 
- Home page mobile
- Home page larger viewports
- All other pages

There is also variations on when the header is sticky on mobile versions and otherwise (mostly tool pages). 

image

image

image

### DIFF
--- a/app/views/shared/_maps_banner.html.erb
+++ b/app/views/shared/_maps_banner.html.erb
@@ -1,0 +1,16 @@
+<div class="l-maps_banner<%= ' l-maps_banner--non-sticky' if hide_sticky_header? %><%= ' l-maps_banner--home' if show_home_banner? %>">
+  <div class="l-constrained">
+    <div class="l-maps_banner__content">
+      <a href="https://moneyandpensionsservice.org.uk/" target="_blank" class="maps_banner__link">
+        <span class="maps_banner__mps-text">
+          <span><%= t('maps_banner.mps_provided_1') %></span>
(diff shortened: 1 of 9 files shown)


app/views/shared/_maps_banner.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <div class="l-maps_banner__content">
+      <a href="https://moneyandpensionsservice.org.uk/" target="_blank" class="maps_banner__link">
+        <span class="maps_banner__mps-text">
+          <span><%= t('maps_banner.mps_provided_1') %></span>
+          <span><%= t('maps_banner.mps_provided_2') %></span>
+        </span>
+
+        <span class="maps_banner__mps-logo"></span>
+
+        <span class="visually-hidden">opens in a new window</span>
```
andreferreiradlw: Why exactly is this span needed?

davidtrussler: For accessibility. Most of our links to external sites have this, so it can be read by screen-readers.

PR #2194, opened 2020-02-17
[DEV APPROVED] Adds three charity logos to introductory section

davidtrussler:
[TP11220](https://maps.tpondemand.com/entity/11220-pace-add-charity-logo-on-top)

This change adds the three charity logos to the top (introductory) section of the PACE page (see screenshot). Additionally it replaces the existing versions of the logos with newer versions in a transparent PNG format since they need to render against a background. We are awaiting SVG versions of these logos. 

image


jordanlg27:
https://maps.tpondemand.com/entity/11220

PR #2201, opened 2020-03-24
11363 sticky banner coronavirus info

davidtrussler:
[TP11363](https://maps.tpondemand.com/entity/11363-sticky-banner-on-mas-website-for)

This work adds a banner to direct users to our articles on Coronavirus. 

There are three PRs that make up the implementation of this component: 
- PR2201 in frontend (this PR) that provides the content and markup for the banner, as well as the Rails functionality to render the component only when this is not recorded in a Cookie as having been previously dismissed, and the reference to the associated Dough component. 
- [PR61](https://github.com/moneyadviceservice/yeast/pull/61) in yeast that provides the styles for the component, and which will be shared with other repos, such as RAD. 
- [PR338](https://github.com/moneyadviceservice/dough/pull/338) in dough that provides the JavaScript component and associated Unit Tests for the behaviour of the banner. This PR also updates the ChatPopup component, necessary to handle the dynamic positioning of the banner on article pages. 

**Banner on the home page**

image

**Banner on an article page**

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem 'whenever', require: false
 # Dependencies
 gem 'adal', git: 'git@github.com:moneyadviceservice/azure-activedirectory-library-for-ruby'
 gem 'cream', '2.1.8'
-gem 'dough-ruby', '~> 5.39'
+gem 'dough-ruby', git: 'git@github.com:moneyadviceservice/dough.git', branch: '11363_Sticky-banner_Coronavirus-info_v5.39'
 gem 'mas-cms-client', '1.20.0'
 gem 'site_search', '0.3.0'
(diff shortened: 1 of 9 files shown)


akenzua:
Looks good. I just think the font color could a bit darker to perfectly match the design.

davidtrussler:
> Looks good. I just think the font color could a bit darker to perfectly match the design. Should it not also reappear on reload if the user click the close icon?

With regard to the colour, I've used the standard MAS palette here, rather than the designs, in order to keep it consistent with the rest of the site. 

On the reload question, the requirement is to not show this again once it is dismissed.

akenzua:
Good then. Approved

PR #2202, opened 2020-04-02
Hides Email Contact Details

andreferreiradlw:
[TP11379](https://maps.tpondemand.com/entity/11379-remove-email-contact-from-mas-footer)

Due to the coronavirus, this PR aims to temporarily hide the email contact information from the last footer panel.

In order to simplify this task, and since all the styles for the footer are in yeast, these temporary styles were added inline in order to provide a quicker implementation and also to make it easier to remove these in the future.


davidtrussler:
👍 - looks good - but only letting inline styles go in this instance since it's a temporary change. It does make it less involved this way.

PR #2203, opened 2020-04-02
updated wpcc dependencies to fix the 404 reset link

kiggundu:
TP-11267: bug fix for wpcc reset generating 404


jordanlg27:
https://maps.tpondemand.com/entity/11267

PR #2216, opened 2020-05-22
[TP-11456] Basic foundation including simple question/answer and tran…

kiggundu:
TP_11456 Simple changes as POC that this solution is viable


jordanlg27:
https://maps.tpondemand.com/entity/11456

PR #2217, opened 2020-06-04
11484 baseline questions view

davidtrussler:
[TP11484](https://maps.tpondemand.com/entity/11484-money-navigator-add-logic-and-styles)

This implements the design and logic of the Questions section of the new Covid-19 Diagnostics tool, based on the [prototype](https://4zy55q.axshare.com/#id=qa62m2&p=start_2).

The tool is located locally at this URL `/en/c19_diagnostics/questionnaire` and is also deployed to UAT [here](https://www.uat.moneyadviceservice.org.uk/en/c19_diagnostics/questionnaire).


davidtrussler:
> Looks good so far David!
> 
> Can you please add label for="questions_Q0_a1" to match the input ID on the radio buttons.
> Can you also please run regular siteimprove accessibility tests. Most of the issues it's complaining about are inherited from the main site but let's try to make the tool as accessible as possible.

Thanks @ladamsatmaps19 - I think the `label` and `input` have matching values throughout, though they are a different format from above. 

On accessibility: happy to do that - I have a list of issues that will need to be done that I am compiling separately and will probably become another card.

ladamsatmaps19:
> > Looks good so far David!
> > Can you please add label for="questions_Q0_a1" to match the input ID on the radio buttons.
> > Can you also please run regular siteimprove accessibility tests. Most of the issues it's complaining about are inherited from the main site but let's try to make the tool as accessible as possible.
> 
> Thanks @ladamsatmaps19 - I think the `label` and `input` have matching values throughout, though they are a different format from above.
> 
> On accessibility: happy to do that - I have a list of issues that will need to be done that I am compiling separately and will probably become another card.

OK, thanks @davidtrussler. The labels look like this in dev tools in the browser. 
<label style="">
                <input type="radio" value="A0" name="questions[Q2]" id="questions_Q2_a0" style="">
                Yes
            </label>

PR #2223, opened 2020-06-18
Feature/tp 11498 c19 hub banner title cta

kiggundu:
TP-11498 banner updates

### DIFF
--- a/app/views/shared/_covid_banner.erb
+++ b/app/views/shared/_covid_banner.erb
@@ -2,7 +2,7 @@
 	<div class="l-constrained">
 		<div class="covid_banner__content">
 			<div class="covid_banner__inner">
-				<a href="#" class="covid_banner__close" data-dough-close>
+				<a href="#" class="covid_banner__close" aria-label=”close” data-dough-close>
 					<svg xmlns="http://www.w3.org/2000/svg" class="svg-icon svg-icon--mobile-close-box">
 						<use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#svg-icon--mobile-close-box"></use>
 					</svg>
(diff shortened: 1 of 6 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11498

app/views/shared/_covid_banner.erb
```diff
@@ -2,12 +2,12 @@
 	<div class="l-constrained">
 		<div class="covid_banner__content">
 			<div class="covid_banner__inner">
-				<a href="#" class="covid_banner__close" data-dough-close>
+				<a href="#" class="covid_banner__close" aria-label=”close” data-dough-close>
```
davidtrussler, line 5: This value should  be available in both languages. It will exist elsewhere on the site. 
(Actually, our usual practice for labelling elements like this is to use text with a class of `visually-hidden`. In that case this would be the same a the mobile menu if that seems like a better option - whatever works).

app/views/shared/_covid_banner.erb
```diff
@@ -2,12 +2,12 @@
...
 					<svg xmlns="http://www.w3.org/2000/svg" class="svg-icon svg-icon--mobile-close-box">
 						<use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#svg-icon--mobile-close-box"></use>
 					</svg>
 				</a>
-				<p class="covid_banner__head"><%= t('covid_banner.head') %></p>
+        <div class="covid_banner__head"><div class="covid_banner__title"> <%= t('covid_banner.head') %> </div> <div class="covid_banner__description">- <%= t('covid_banner.description') %></div></div>
```
davidtrussler, line 10: Probably better to keep the `<p>` tag here and add the inner content as `<span>` elements if it's just a case of adding a bold style. 

Also, the increased volume of content means this is breaking the layout on smaller viewports. Might need to see if the text size can be reduced or the copy edited.

PR #2225, opened 2020-06-23
Feature/tp 11520 redirect hub url

kiggundu:
This redirects the existing c19 hub url to the new one as per TP-11520


jordanlg27:
https://maps.tpondemand.com/entity/11520

PR #2230, opened 2020-06-30
Money navigator results add print option

davidtrussler:
[TP11446](https://maps.tpondemand.com/entity/11446-moneynavigator-add-print-option-to-results)

This work adds the functionality to allow the user to print the contents of the results page once they have used the tool.

The design on which this is based is [here](https://4zy55q.axshare.com/#id=8c2jys&p=results_page&g=12&view=default). 

The URL for the tool on frontend is `/en/tools/money-navigator-tool`. 

There already exists a site-wide print style-sheet; this work makes a minor update to that and adds more specific styles for this tool. 

image

### DIFF
--- a/app/assets/javascripts/components/MoneyNavigatorResults.js
+++ b/app/assets/javascripts/components/MoneyNavigatorResults.js
@@ -1,16 +1,21 @@
 define(['jquery', 'DoughBaseComponent', 'utilities'], function($, DoughBaseComponent, utilities) {
   'use strict';
 
-  var MoneyNavigatorResults;
-  
+  var MoneyNavigatorResults,
+      i18nStrings = {
+        printBtnText: 'Print this'
+      };
+
   MoneyNavigatorResults = function($el, config) {
     MoneyNavigatorResults.baseConstructor.call(this, $el, config);
 
+    this.i18nStrings = (config && config.i18nStrings) ? config.i18nStrings : i18nStrings;
     this.$headingContent = this.$el.find('[data-heading-content]'); 
     this.$sections = this.$el.find('[data-section]'); 
     this.$sectionTitles = this.$el.find('[data-section-title]'); 
     this.$headingTitles = this.$el.find('[data-heading-title]'); 
     this.$overlay = this.$el.find('[data-overlay]'); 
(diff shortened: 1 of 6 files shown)


akenzua:
I do not know how far this work has gone but i saw a bunch of warning in the console, mostly semantics and two particular warnings about ids appearing more than once on a page. 
money_helper_warnings

davidtrussler:
> I do not know how far this work has gone but i saw a bunch of warnings in the console, mostly about semantics and two particular warnings about ids appearing more than once on a page. I can see all other issues on on TP cards.

@akenzua Yes, those have been around for a while and mostly appear to reference some earlier work so i would consider them to be outside the scope of this PR. They also appear not to be fired outside of a local environment. 
The ones about multiple IDs are more relevant to this though, but I'm unable to see those myself. Do you have a specific journey that triggers these?

akenzua:
> > I do not know how far this work has gone but i saw a bunch of warnings in the console, mostly about semantics and two particular warnings about ids appearing more than once on a page. I can see all other issues on on TP cards.
> 
> @akenzua Yes, those have been around for a while and mostly appear to reference some earlier work so i would consider them to be outside the scope of this PR. They also appear not to be fired outside of a local environment.
> The ones about multiple IDs are more relevant to this though, but I'm unable to see those myself. Do you have a specific journey that triggers these?

Nothing in particular.  I just clicked through to the result page.

davidtrussler:
> > > I do not know how far this work has gone but i saw a bunch of warnings in the console, mostly about semantics and two particular warnings about ids appearing more than once on a page. I can see all other issues on on TP cards.
> > 
> > 
> > @akenzua Yes, those have been around for a while and mostly appear to reference some earlier work so i would consider them to be outside the scope of this PR. They also appear not to be fired outside of a local environment.
> > The ones about multiple IDs are more relevant to this though, but I'm unable to see those myself. Do you have a specific journey that triggers these?
> 
> Nothing in particular. I just clicked through to the result page.

@akenzua OK, I have found those warnings. They are coming from content that is loaded from the CMS so those will also be out of scope for this PR. Can you see if there are any other issues or if it can be approved?

PR #2238, opened 2020-07-15
Feature/tp 11557 c19 hub money navigator

kiggundu:
This is PR is to deliver [TP-11557](https://maps.tpondemand.com/RestUI/Board.aspx#page=userstory/11557&appConfig=eyJhY2lkIjoiMzg2RDNFNEYzQ0UyRTIxMkZDMEY3QTNBNkFBOEFGQjEifQ==). Details and screenshots can be found in the ticket.


jordanlg27:
https://maps.tpondemand.com/entity/11557

app/assets/stylesheets/components/page_specific/_content_hub.scss
```diff
@@ -36,15 +36,14 @@
 .content-hub__hero-image {
   padding-bottom: 100%;
```
davidtrussler, line 37: I would suggest making this percentage value equal to the aspect ration of the image so that it fits the space exactly. In this case that can be `250 / 479 * 100%`. This helps with the positioning of the elements that follow in the flow.

app/assets/stylesheets/layout/page_specific/_content_hub.scss
```diff
@@ -71,17 +74,24 @@ $content-hub-two-col-breakpoint: $content-hub-width-three-module + px($gutter-wi
...
+  @include respond-to($content-hub-two-col-breakpoint) {
+    @include clearfix();
+    min-height: $content-hub-width-one-module*1.7;
+  }
 }
 
 .l-content-hub__hero-content {
   padding: 0 $content-hub-horizontal-space;
+  position: absolute;
+  top: 50%;
```
davidtrussler, line 90: If the image is treated as above then this need not be absolutely positioned and these two lines can be removed.

app/assets/stylesheets/components/page_specific/_content_hub.scss
```diff
@@ -36,15 +36,14 @@
...
+  width: 100%;
 
   @include respond-to($content-hub-one-col-breakpoint) {
     position: absolute;
     top: 0;
     bottom: 0;
     padding-bottom: 0;
-    width: 50%;
     padding: 0 $content-hub-horizontal-space;
   }
```
davidtrussler, line 48: The change above should also mean this separate style for the larger viewport isn't needed.

davidtrussler:
I'm leaving a few comments on specific elements for this but it occurs to me that the layout has changed in the hero section from what it was previously. 

Ideally we wouldn't attempt to change the layout of what is supposed to be a very generic container designed for varied content instances. The result of doing this is that the layout and other hub pages that already exist will break.

If it is the case that this more generic layout needs to remain applicable to other existing instances of the hub page, we may need to add some styles specific to this particular page to effect the required layout changes. But ideally we would retain the layout and fit new content to it rather than change the layout to fit new content. 

This isn't really a dev matter but rather one for a project or delivery manager.

PR #2239, opened 2020-07-20
[TP11626] Updates CSS on Urgent Actions for amended markup from CMS

davidtrussler:
[TP11626](https://maps.tpondemand.com/entity/11626-money-navigator-urgent-ctas-improvements)

When the Money Navigator tool was first released we dealt with the layout of the `Urgent Actions` section on the results page by tightly coupling the styles for the eight possible variations of the content of these sections with their content. This led to a very brittle layout whereby small changes in content could (and did) break the layout. 

This work relies on the addition of some class names to the markup in the CMS that makes the CSS able to render the correct layout in a more general way and dispenses with the selectors that are overly specific. 

The layout of these sections should remain largely unchanged as a result of this work. 

An example of such a callout is shown tin the screenshot below.

image

### DIFF
--- a/app/assets/stylesheets/layout/page_specific/_money_navigator_results.scss
+++ b/app/assets/stylesheets/layout/page_specific/_money_navigator_results.scss
@@ -23,6 +23,12 @@ $transition-time: 0.4s;
 		.urgent-actions__actions {
 			@include column(12); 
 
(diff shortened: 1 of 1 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11626

PR #2242, opened 2020-07-21
[TP11636] Money Navigator: updates completed messages

davidtrussler:
[TP11636](https://maps.tpondemand.com/entity/11636-money-navigator-add-translations-for-question)

This work updates the completion messages displayed for each question by passing the copy to the JS component via the `dough-config` attribute. When the other text for the page was implemented this instance was found to be problematic due to differing constructions in the Welsh and English languages. The decision was taken to update the order of the display of the message in English to match that in Welsh so that both are syntactically correct. 

Specifically, it
- Adds the copy to the translation files
- Adds references to the messages in the `dough_money_navigator_questions_config` attribute in the view.  
- Updates the word order of the message in the JS component
- Adds the completion message to the JS component

**Current (English)**

image

**Current (Welsh)**

image

**Updated (English)**

image

**Updated (Welsh)**

image


jordanlg27:
https://maps.tpondemand.com/entity/11636

PR #2248, opened 2020-07-29
11623 money navigator accessibility keyboard tabbing improvements

ladamsatmaps19:
[TP11623](https://maps.tpondemand.com/RestUI/Board.aspx#page=board/5682182231901749200&appConfig=eyJhY2lkIjoiNjA5RjRCQzM4NTlDRjBBOEQ4QzM2ODUxMEVENDhFNkUifQ==&boardPopup=userstory/11623/silent)

This PR covers the fix for the new UI Yes/No buttons by restructuring the CSS so that users can navigate between the two buttons using keyboard and screen reader. 

image

Part of this ticket was not completed but I have created a new ticket for the back button issue as it should be a ticket on it's own.


app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
```diff
@@ -139,37 +140,101 @@
...
 				}
+
 			}
 
-			.button--yes {
-				margin-right: 0;
+			.button--yes{
+				[type="checkbox"]:not(:checked) + label:before,
+				[type="checkbox"]:checked + label:before {
+					content:'Yes';
```
davidtrussler, line 214: The problem with using generated content for this is that we lose the translations. It would be possible to do this by making the content conditional on the `theme-cy` class that is attached in the `<body>` tag but really this content should be generated from the translation file.

davidtrussler:
I've left a comment about the problems of using generated content in this context. I think there is a simpler solution though that applies the focus style to the label that is the sibling of the hidden input that receives focus. That way we can leave the structure as it is.

app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
ladamsatmaps19: Ah yes, I haven't worked with translation files @davidtrussler.

ladamsatmaps19: Happy for @davidtrussler to use his suggested approach as long as input is not hidden and is keyboard/screen reader navigable.

ladamsatmaps19: I've reworked the buttons so that they use the translation <span> rather than the Yes/No coming from the CSS. totally my bad on that!

ladamsatmaps19: & my git command line captures spelling mistakes then uses them later. I've cleared the cache but sorry for the comments!

davidtrussler:
@ladamsatmaps19 Hmm, looks like the style for the checked input is extending to the other checkboxes as it stands: 

image

I'm still thinking we can achieve this more simply though. Where we have the rule below for changing the style on checking an input, could we use something similar similar to add the focus style to the label when the associated input has focus?

image

ladamsatmaps19:
Thanks @davidtrussler css lines moved to make them specific to the yes/no buttons

davidtrussler:
This works much better. 🙂 

I'm still thinking it is overcomplicating the solution though. My thinking on this was to add to the `.response__control` element something like below so that when that element gains focus the label (`.response__text`) displays the visible focus style. Because the input is using our generic `.visually-hidden` class it is not visible but is still part of the document flow and useable by assistive technologies. 

image

ladamsatmaps19:
OK @davidtrussler I'll leave it with you.

davidtrussler (approved):
LGTM! 👍

PR #2252, opened 2020-08-12
[TP-11691] Updates global nav style to fix knock-on problem on DAL

davidtrussler:
[TP-11691](https://maps.tpondemand.com/entity/11691-bug-dalt-current-location-warning-message)

This update moves a `is-hidden` style in the global navigation to make it more specific to that component. This has recently been updated to make it less generic which in turn caused a problem with the error message that is shown on the Debt Advice Locator when the user's device is unable to get a location. 

The `is-hidden` style that is used both here and on the DAL tool is now a new style created in Dough. 

The change in the DAL repo that was required to be made is in [PR172](https://github.com/moneyadviceservice/debt-advice-locator/pull/172). 

The change in Dough that was required to be made is in [PR 340](https://github.com/moneyadviceservice/dough/pull/340).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem 'whenever', require: false
 # Dependencies
 gem 'adal', git: 'git@github.com:moneyadviceservice/azure-activedirectory-library-for-ruby'
 gem 'cream', '2.1.8'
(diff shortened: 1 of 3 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11691

PR #2254, opened 2020-08-25
[TP-11680] Fixes failing Unit test in `MoneyNavigatorQuestions_spec.js`

davidtrussler:
[TP-11680](https://maps.tpondemand.com/entity/11680-money-navigator-technical-debt-fix-failing)

This is part of the Technical Debt arising from the work on the Money Navigator tool. It fixes a Unit Test that was broken on the deployed version of the tool.

The test checks for the display state of the banner that appears at the start of the questions (see screenshot below) and is only required to be shown for the first question. 

The problem was that the test runner was unable to locate the element in the fixture because it is located outside of the component under test. This change makes the element available to the test. 

It also makes the element location within the DOM more specific in the component itself. That change is made more for consistency than necessity. 

image


jordanlg27:
https://maps.tpondemand.com/entity/11680

ladamsatmaps19 (approved):
Code looks good to me. Can @kiggundu please take a look when you have a chance?

PR #2256, opened 2020-08-26
[TP-11721] Add SameSite=None whenever we have X-Syndicated-Tool header

ryanharkins:
Using middleware instead of overriding `set_cookie_header` in `/action_dispatch/middleware/cookies.rb`. This way we can set it by default for syndicated tools only and the other parts we leave as is.


jordanlg27:
https://maps.tpondemand.com/entity/11721

kiggundu (approved):
LGTM from a code perspective
Pre-production QA testing will confirm this fix adequately addresses the issue in practice.

PR #2257, opened 2020-08-26
[TP-11682] Fixes some small UI inconsistencies

davidtrussler:
[TP-11682](https://maps.tpondemand.com/entity/11682-money-navigator-technical-debt-fix-small)

This work fixes some small UI issues on the Money Navigator Questions page. These concern the `Continue` and `Back` buttons as they render on smaller viewports (see screenshots below). The problem is that their common container is responding wrongly to the viewport width. The solution is to change the width of the container at a wider breakpoint to allow the buttons to display at the same width as other elements. 

image

image


jordanlg27:
https://maps.tpondemand.com/entity/11682

ladamsatmaps19 (approved):
Code looks good to me. @akenzua could you please test this in the browser?

PR #2273, opened 2020-09-29
[DEV APPROVED] Adds `f.errors_for` & `f.form_row` tags

davidtrussler:
[TP11594](https://maps.tpondemand.com/entity/11594-money-navigator-error-anchor-links-dont)

This work addresses a bug in the Money Manager tool whereby any validation errors that are caught server-side are not properly reported to the user. The main effect of this is that the tool correctly displays a generic message but the links within that do not take the user to the appropriate question and that question is not highlighted as containing an error. 

image

**Example: the message above is displayed but the link `Q12 can't be blank` does not take the user to the question containing the error.** 

This is caused by an incomplete implementation of the Dough Form Builder methods. Specifically we are missing the required `f.form_row` and `f.errors_for` tags which display the question with an `is-errored` class and add the appropriate id to the question. 

image

**Example: the question above has the correct tags added.**

### DIFF
--- a/app/views/money_navigator_tool/questionnaire.html.erb
+++ b/app/views/money_navigator_tool/questionnaire.html.erb
@@ -37,78 +37,82 @@
                 data-question-grouped-group-titles="<%= question[:group_titles] %>"
               <% end %>
             >
-              <% if index > 0 %>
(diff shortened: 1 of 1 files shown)


andreferreiradlw (approved):
LGTM

PR #2281, opened 2020-10-14
Feature/tp 11798 stepchange update

kiggundu:
This change delivers [TP-11798](https://maps.tpondemand.com/entity/11807-money-navigator-important-update-to-urgent) as part of the stepchange updates

### DIFF
--- a/app/models/concerns/money_navigator/rules/urgent_action.rb
+++ b/app/models/concerns/money_navigator/rules/urgent_action.rb
@@ -9,14 +9,14 @@ class MoneyNavigator::Rules::UrgentAction
 
         {
           # Get Free debt advice now' Rules
-          # Any of these Q4A1, Q6A6, Q7A1-A9, Q10A3 PLUS the regional variation
+          # Any of these Q4A1, Q4A2, Q6A4, Q6A5, Q6A6, Q7A1-A9  BUT NOT IF HAVE SELECTED Q10A1 PLUS the regional variation
           heading_code: 'H1',
           content_rules: [
             {
               triggers: [
                 { q0: 'a1' },
-                { q4: 'a1', q6: ['a6'], q7: %w[a1 a2 a3 a4 a5 a6 a7 a8 a9], q10: 'a3' },
-                { q1: 'a2' }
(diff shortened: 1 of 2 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11798

app/models/concerns/money_navigator/rules/urgent_action.rb
```diff
@@ -9,41 +9,41 @@ def self.all
...
-          # Any of these Q4A1, Q6A6, Q7A1-A9, Q10A3 PLUS the regional variation
+          # Any of these Q4A1, Q4A2, Q6A4, Q6A5, Q6A6, Q7A1-A9  BUT NOT IF HAVE SELECTED Q10A1 PLUS the regional variation
           heading_code: 'H1',
           content_rules: [
             {
               triggers: [
                 { q0: 'a1' },
-                { q4: 'a1', q6: ['a6'], q7: %w[a1 a2 a3 a4 a5 a6 a7 a8 a9], q10: 'a3' },
-                { q1: 'a2' }
+                { q4: %w[a1 a2], q6: %[a4 a5 a6], q7: %w[a1 a2 a3 a4 a5 a6 a7 a8 a9]},
```
davidtrussler, line 18: I think this should be `%w[a4 a5 a6]` (or commas as array separators). And likewise in the three instances below.

kiggundu: Ahhh.... good spot... looks like it might have come from a merge... 
not sure how the tests missed that as well... looking into it

davidtrussler: > Ahhh.... good spot... looks like it might have come from a merge...
> not sure how the tests missed that as well... looking into it
> 
Yeah I ran the tests with that as it is and changed to `%w` and it didn't make any difference, which is a bit odd.

kiggundu: The tests rightly passed because that rule contains q4 and q7 answers that were being provided. Reworking the testing to cover such an edge case would be quite involving (i.e. another 2 days of effort at least) so i've chosen to simply fix the mistake.

davidtrussler: Yeah, that seems fair enough!

davidtrussler (approved):
LGTM! 👍

PR #2288, opened 2020-11-11
Money navigator, font inconsistencies on the Results page overlays

davidtrussler:
[TP11561](https://maps.tpondemand.com/entity/11561-review-font-sizes-in-action-callout)

There are a few inconsistencies in the rendering of text on the Results page of the Money Navigator tool when the user opens the overlays to view content. This is caused by the content being fetched from the CMS and thereby missing styles that would be available in the context of an article page. This work adds the missing styles to the Money Navigator stylesheet so that the display is consistent with the rest of the site.

Additionally, the overlay is updated to present its content within an `<article>` tag, which allows the content to be rendered more semantically and thereby better in terms of accessibility. 

The main visual differences are summarised below. 

The overlay heading becomes an `<h1>` within an `<article>` element that is better semantically and is aligned with a MAS article page
| Current | Updated |
|---|---|
| image | image |

The standard font size is updated to be consistent with a MAS article page
| Current | Updated |
|---|---|
| image | image |

The font sizing is made consistent within the `add-action` components
| Current | Updated |
|---|---|
| image | image |

Lists are styled to be consistent with a MAS article page
| Current | Updated |
|---|---|
| image | image |

### DIFF
--- a/app/assets/stylesheets/layout/page_specific/_money_navigator_results.scss
+++ b/app/assets/stylesheets/layout/page_specific/_money_navigator_results.scss
@@ -291,7 +291,6 @@ $transition-time: 0.4s;
 				margin-bottom: 0; 
 
 				p, a {
-					font-size: 0.875rem; 
 					margin: $baseline-unit 0;
 				}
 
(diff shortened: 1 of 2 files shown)


andreferreiradlw (approved):

ladamsatmaps19 (approved):
Nice!

PR #2295, opened 2021-01-06
TP 11926 mn update urgent debt advice logic

thenapking:
[TP 11926](https://maps.tpondemand.com/entity/11926-money-navigator-update-urgent-debt-advice)

Alter logic and test such that when the user is not self-employed and any answer except No is given for Q6, Q7 Q9, the Urgent Debt Advice banner is triggered
Add a CSS class to allow hiding of individual answers
Hide the answer to Q3 which allows users to select "I am not self-employed" even if they have already said they are self employed in a previous question.

### DIFF
--- a/app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
+++ b/app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
@@ -133,9 +133,6 @@
 		}
 	}
 
-	.question__response__hidden {
-		display: none;
-	}
 }
 
 // Component initialised
@@ -145,6 +142,10 @@
 
 		&.question--active {
 			display: block; 
+		}
+
+		.question__response--hidden {
+			display: none;
 		}
 
 		&[data-question-multiple] {
@@ -221,7 +222,7 @@
 						outline:none;
 					}
 				}
-	
+
 				.response__control {
 					&:checked + .response__text {
 						color: $color-white; 
--- a/app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
+++ b/app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
@@ -132,6 +132,10 @@
 			padding: 0 0 0 $baseline-unit * 4;
 		}
 	}
+
+	.question__response__hidden {
+		display: none;
+	}
 }
 
 // Component initialised
--- a/config/locales/money_navigator_tool/money_navigator_tool.en.yml
+++ b/config/locales/money_navigator_tool/money_navigator_tool.en.yml
@@ -136,6 +136,7 @@ en:
             text: 'I am not self employed'
             default: true
             sort_order: 1
+            hidden: true
       - code: Q4
         type: radio
         custom: true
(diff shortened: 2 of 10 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11926

config/locales/money_navigator_tool/money_navigator_tool.en.yml
```diff
@@ -136,6 +136,7 @@ en:
             text: 'I am not self employed'
             default: true
             sort_order: 1
+            hidden: true
```
davidtrussler, line 139: We shouldn't hide this: it is still necessary if the user has indicated they are not self employed in an earlier question. Additionally any display logic like this appears to be outside the scope of this TP card so I would remove that from this work and deal with it separately if necessary (which I do not believe it is).

thenapking: Hey David, As I understand it this question is _only_ displayed when the user has selected that they are self-employed. The question text reads "If you are self-employed do you know your income". I am not self employed is automatically selected as the default, so it will be selected even if the user doesn't answer this, which is what happens if they select any other answer to the previous question. 
The reason this has been added into this update is that on asking for clarification of the requested changes to the logic, a complicated change was requested in the scenario that the user has told us they are self-employed in the previous question, and then told us they aren't in this question.  To avoid this unnecessary complication we should not give the users the ability to select contradictory answers.

davidtrussler: Yes, though that is question is excluded dynamically. If you load the tool without JS for instance you have all the questions. In this case it would be possible to answer that you are not self-employed but still be forced to answer Yes/No to the question that is only relevant to self-employed people. That is why the default is set to not self-employed. This consideration could also apply to some assistive technologies.

app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
```diff
@@ -132,6 +132,10 @@
 			padding: 0 0 0 $baseline-unit * 4;
 		}
 	}
+
+	.question__response__hidden {
+		display: none;
+	}
```
davidtrussler, line 138: Could we move this rule to be within the `[data-dough-money-navigator-questions-initialised="yes"]` section of the CSS (below)? That way this will only apply to the enhanced display of the questionnaire. Probably right at the top would be fine since it is a useful general rule.

app/assets/stylesheets/layout/page_specific/_money_navigator_questionnaire.scss
```diff
@@ -132,6 +132,10 @@
 			padding: 0 0 0 $baseline-unit * 4;
 		}
 	}
+
+	.question__response__hidden {
```
davidtrussler, line 136: Could we rename this to `question__response--hidden` which more closely follows our BEM convention?

davidtrussler (approved):
LGTM! 👍

thenapking:
Merged in [TP-11921](https://github.com/moneyadviceservice/frontend/pull/2296)

PR #2299, opened 2021-01-20
TP-11667 Money Navigator A11Y back button

davidtrussler:
[TP-11667](https://maps.tpondemand.com/entity/11667-money-navigator-accessibility-on-back-button)

This work updates the logic that deals with the display of individual questions on the Money Navigator tool as the user navigates through the questions by keyboard via the `back` button. It ensures that focus is moved to the top of the previous question so that the flow for the keyboard user remains intuitive.

It also adds a Unit Tests for the new functionality which in turn required an update to the existing tests for the component because a clash occurred with other tests when trying to obtain a value for `document.activeElement`. This update in turn required a small change in an existing test. All now run OK. 

| When the back button on a question is activated | Focus is moved to the first input element of the previous question |
| :--- | :--- |
| image | image |

### DIFF
--- a/app/assets/javascripts/components/MoneyNavigatorQuestions.js
+++ b/app/assets/javascripts/components/MoneyNavigatorQuestions.js
@@ -512,6 +512,12 @@ define(['jquery', 'DoughBaseComponent'], function ($, DoughBaseComponent) {
       .find('.question__counter')
       .text(progress + '% ' + this.i18nStrings.messages.completed);
 
+    if (dir === 'prev') {
+      if (activeIndex >= 0) {
+        questions[activeIndex].focus(); 
+      }
+    }
+
     if (activeIndex == 0) {
       this.$banner.removeClass(
         'l-money_navigator__banner' + '--' + this.hiddenClass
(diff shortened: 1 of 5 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11667

ladamsatmaps19:
Hi @davidtrussler, am testing this but when I use back button and go back to previous page the selected radio isn't the first answer? this PR reads as though that is the expectation?

davidtrussler:
> Hi @davidtrussler, am testing this but when I use back button and go back to previous page the selected radio isn't the first answer? this PR reads as though that is the expectation?

This just changes the focus, it shouldn't affect which input is selected. Saying that, to meet the Acceptance Criteria in the TP card the focus should be on the parent element of the question. I'm going to do some more work on this and return it to PR status later.

davidtrussler:
> > Hi @davidtrussler, am testing this but when I use back button and go back to previous page the selected radio isn't the first answer? this PR reads as though that is the expectation?
> 
> This just changes the focus, it shouldn't affect which input is selected. Saying that, to meet the Acceptance Criteria in the TP card the focus should be on the parent element of the question. I'm going to do some more work on this and return it to PR status later.

@ladamsatmaps19 I've added a new [commit](https://github.com/moneyadviceservice/frontend/pull/2299/commits/3bcb7c1f6442170bf0ab130484af57d5d8ac9d63) to change the element that receives focus to be the parent element of the question. I believe this is closer to the Acceptance Criteria.

ladamsatmaps19 (approved):
I think this fulfils the acceptance criteria of this ticket. Thanks for revising @davidtrussler

PR #2302, opened 2021-01-22
[Bugfix] Resolve issue with glibc and nokogiri

baphled:
Addresses issue with Nokogiri not being compatible with CI, also pins versions to make sure we're not accidentally updating gems by mistake.

This was introduced when we accidentally upgraded FactoryBot to 5.2.0, from 4.10.0

The change pins FactoryBot to the previous version, so that we only update FactoryBot when we're ready to update the factory files.

The upgrade of FactoryBot happened here: fa628fa#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fR508


thenapking:
This looks good to me.  The master branch has Nokogiri locked at 1.10.10 and I have seen the same build error on Jenkins with another project where I tried to move beyond that version.  Ideally though you would revert all gems that have changed on this branch to the versions currently in use by the master branch, unless your code update relies specifically on features from the newer versions.

benlovell:
Closed by #2354.

PR #2304, opened 2021-01-26
TP 11959 Randomise Url on PACE Microsite

thenapking:
[TP 11959](https://maps.tpondemand.com/entity/11959-urgent-randomise-url-on-pace-microsite)

This updates behaviour on /moneyadvisernetwork/online.  Note that the buttons on /moneyadvisernetwork all lead to the online route.  This route is also referenced by other pages, so redirecting must be handled from there.

In PACE controller
-add array of redirect urls
-set a cookie for index in array of chosen site 

In online view, add meta tag for delayed redirect

### DIFF
--- a/app/views/pace/online.html.erb
+++ b/app/views/pace/online.html.erb
@@ -1,3 +1,5 @@
+<% set_meta_tags(refresh: @redirect_string) %>
+
 <style>
   .no-js .js-only-text { display: none; }
 </style>
--- a/app/views/pace/online.html.erb
+++ b/app/views/pace/online.html.erb
@@ -1,17 +1,10 @@
 <% set_meta_tags(refresh: @redirect_string) %>
 
-<style>
-  .no-js .js-only-text { display: none; }
-</style>
-
 <br />
 <br />
 <center>
   <strong>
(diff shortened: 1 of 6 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11959

davidtrussler:
The Acceptance Criteria in the TP Card is unclear - I have added a comment to that to get some clarification - but I have a couple of general comments based on this lack of clarity: 
- We currently display text on the page to which the user if forwarded to state that JavaScript needs to be enabled. As I understand it the changes in this work make this incorrect so we should remove it and consider an alternative. 
- I'm unclear on the need to set a delay in the redirect.

app/views/pace/online.html.erb
```diff
@@ -1,3 +1,5 @@
+<% set_meta_tags(refresh: @redirect_string) %>
```
davidtrussler, line 1: I'm finding that this appends the redirect string to the existing URL rather than replaces it when I run it locally. I'm not sure if that is only a problem related to the environment.

davidtrussler (approved):
LGTM with most recent changes. 👍

PR #2305, opened 2021-02-01
11822 money navigator grouped questions focus issues

davidtrussler:
[TP-11822](https://maps.tpondemand.com/entity/11822-money-navigator-focus-styles-on-grouped)

This work improves the experience for a keyboard user on the enhanced UI for the 2 stage question ("Are you behind with rent or mortgage payments?"). This currently does not properly display focus styles and breaks badly on the second stage (see screenshots below). 

| Initial stage | Second stage |
| --- | --- |
| image | image |

The solution is to refactor the markup a little to aid the natural keyboard navigation and to dynamically set focus styles on keyboard events to allow navigation by a combination of the tab and arrow keys based on this WebAim [Keyboard Accessibility](https://webaim.org/techniques/keyboard/) specification.

### DIFF
--- a/app/assets/javascripts/components/MoneyNavigatorQuestions.js
+++ b/app/assets/javascripts/components/MoneyNavigatorQuestions.js
@@ -56,6 +56,7 @@ define(['jquery', 'DoughBaseComponent'], function ($, DoughBaseComponent) {
     this._updateDOM(this.dataLayer); 
     this._setUpMultipleQuestions(); 
     this._setUpGroupedQuestions(); 
+    this._setUpKeyboardEvents(); 
     this._setUpJourneyLogic(); 
     this._initialisedSuccess(initialised);
   };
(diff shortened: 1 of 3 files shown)


ladamsatmaps19:
Hi @davidtrussler when I click on the other options, other than no, the next screen doesn't load. Comparison with live tool screenshots attached. This is happening when navigating with mouse or keyboard. 

Screenshot 2021-02-02 at 11 22 22
Screenshot 2021-02-02 at 11 21 59

davidtrussler:
> Hi @davidtrussler when I click on the other options, other than no, the next screen doesn't load. Comparison with live tool screenshots attached. This is happening when navigating with mouse or keyboard.
> 
@ladamsatmaps19 This is caused by inconsistent padding values on the parent element of the question at different widths. I have kept these the same which makes a small (but hopefully insignificant) change to the layout but fixes this issue.

ladamsatmaps19 (approved):
Looks good to me! nice work @davidtrussler

PR #2306, opened 2021-02-04
Tp 11967 money navigator keyboard nav

davidtrussler:
[TP-11967](https://maps.tpondemand.com/entity/11967-money-navigator-a11y-question-3-options)

In a previous change to the UI on the question "If you are self-employed, do you know what your monthly income is now?" (accessed only by users who have previously indicated that their employment status is self-employed) we removed the option "I am not self employed" from the display so avoid slightly contradictory messaging. Because that option remains checked by default this causes a scenario whereby a keyboard user is unable to navigate to the inputs that remain displayed.

The agreed solution is to set a checked state on the next sibling of any input that is both checked by default and hidden. 

| The input that is checked by default is hidden | This means the remaining inputs cannot be accessed by keyboard users | The default input remains hidden and a checked state is added to its next sibling |
|---|---|---|
| image | image | image |

### DIFF
--- a/app/assets/javascripts/components/MoneyNavigatorQuestions.js
+++ b/app/assets/javascripts/components/MoneyNavigatorQuestions.js
@@ -539,6 +539,20 @@ define(['jquery', 'DoughBaseComponent'], function ($, DoughBaseComponent) {
           .find('.question__content')
           .append(div);
       }
+
+      // Move default input if it is also hidden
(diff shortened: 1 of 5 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/11967

ladamsatmaps19 (approved):
Works for me!

PR #2312, opened 2021-02-25
12012 civic cookie implementation

davidtrussler:
[TP-12012](https://maps.tpondemand.com/entity/12012-mas-cookie-banner-changes)

This work removes the current cookie implementation on MAS and replaces it with a third party solution. The steps taken to do this are: 
- Adds the script that calls the Civic CDN to the site <head> tag
- Adds a new MAS module to implement the Civic component
- Adds a stylesheet for the component to apply MAS styles

There remains some work still to do and this will be logged as separate Technical Debt tasks: 
- Write unit tests for module
- Fully remove the current cookie implementation
- Fix the issue identified with the locale - the plan was to use the existing MAS i18n module but this is not working correctly so a temporary solution has been implemented here. 
- Make some updates for a11y: although the solution does not introduce any A11Y issues there are some areas where this is not using the MAS standards (underlines on links for example). These should be amended to provide the user with an experience consistent with the rest of the site.

The implementation above adds a panel that is opened by clicking on the "C" icon shown below, located at the bottom of the viewport. 

image

### DIFF
--- a/app/assets/javascripts/modules/mas_cookieController.js
+++ b/app/assets/javascripts/modules/mas_cookieController.js
@@ -1,0 +1,142 @@
+var defaults = {
+  theme: 'light', 
+  optionalCookies: [],
+  text: {}, 
+  branding: {}
+};
+
+var CookieController = function (opts) {
+  this.config = Object.assign(defaults, opts);
+
+  // TODO: replace locale and textStrings with the existing i18n module
+  // This appears to be not functioning as it should right now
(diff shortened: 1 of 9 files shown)


andreferreiradlw (approved):
Happy to approve this since there's still a lot of work to be done mentioned on the PR description. It works fine for me locally.

ladamsatmaps19 (approved):
Looks good to me!

PR #2333, opened 2021-04-27
[TP-12233] Applies fix for DAL search by town not working

davidtrussler:
[TP-12233](https://maps.tpondemand.com/entity/12233-dalt-search-by-town-not-working)

This adds a reference to the cookie `__zjc*` to the Civic `necessaryCookies` array. 

This is an attempt to fix a problem on the Debt Advice Locator whereby the Search feature sometimes fails. It seems that this cookie is being set to expire immediately by a script that is being injected into the DOM. Although the cause of this is not clear it seems likely that it is because this cookie is not allowed by the Civic implementation. 

This may not fix the problem but we have agreed to try this to observe the effect on this tool in the live environment.

### DIFF
--- a/app/assets/javascripts/modules/mas_cookieController.js
+++ b/app/assets/javascripts/modules/mas_cookieController.js
@@ -90,7 +90,12 @@ var CookieController = function (opts) {
 }
 
 CookieController.prototype.addNecessaryCookies = function() {
-  this.config.necessaryCookies = ['action-plan-*', '_iz_uh_ps_', '_iz_sd_ss_'];
+  this.config.necessaryCookies = [
+    'action-plan-*', // Redundancy Pay Calculator
+    '__zjc*',        // Debt Advice Locator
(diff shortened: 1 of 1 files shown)


jordanlg27:
https://maps.tpondemand.com/entity/12233

thenapking:
When I run this locally search still doesn't appear to work, but I agree we should deploy this to a production environment.  I also can't find the referenced version of dough in the gem file.  Other than that it is fine.

davidtrussler:
> When I run this locally search still doesn't appear to work, but I agree we should deploy this to a production environment. I also can't find the referenced version of dough in the gem file. Other than that it is fine.

Yeah the search doesn't return results locally but it does at least return a meaningful message. On production it (sometimes) returns an odd script that appears to set a cookie to expired. I don't think any harm will come from deploying it even if it doesn't fix the problem.

PR #2336, opened 2021-05-13
TP-12090 Remove Cream 

thenapking:
[TP-12090](https://maps.tpondemand.com/entity/12090-email-feature-does-not-work)

CRM is no longer in use and contains no valid information
DelayedJob failed to send emails when attempting to fetch user information using Cream
Remove ActiveRecord hooks to communicate with Cream when user created/updated
Relevant tests


jordanlg27:
https://maps.tpondemand.com/entity/12090

davidtrussler (approved):
This looks good to me, though ideally we'd want to a Ruby developer to look over it as well but 🤷

PR #2337, opened 2021-05-13
TP 12557 XSS in DALT

thenapking:
[TP-12557](https://maps.tpondemand.com/entity/12557-fix-xss-cross-site-scripting-issue)

Add test for OrganisationSearchDecorator, which includes a XSS attack in the query string.
Update OrganisationSearchDecorator to strip <script> tags and sanitize any query string, before rendering as html_safe


jordanlg27:
https://maps.tpondemand.com/entity/122557

PR #2343, opened 2021-05-26
TP 12098 update the password reset email

thenapking:
Update send from email


jordanlg27:
https://maps.tpondemand.com/entity/12098

PR #2348, opened 2021-06-17
12553 money navigator scroll on mobile

davidtrussler:
[TP-12553](https://maps.tpondemand.com/entity/12553-money-navigator-unable-to-fully-scroll)

This work addresses the problem on iOS devices that are unable to switch between scrolling contexts when rendering content within an iFrame. This is specific to the content in expandable sections on the results page of the money Navigator tool, activated by clicking on the headers such as "Getting back on track after a severe income drop" shown below. 

image

The solution adopted here is to dynamically resize the body element when the content is expanded so that it does not need to be scrolled. With this in place users of devices affected do not need to change their scrolling context and area able to view the full content. 

It is not a perfect solution and some other issues are raised by it: 
- the content cuts off when the page is resized
- the expanded content is short of the full page height when the height of the content is less than the height of containing page

There is a partial fix for the second of these issues in this PR but this will need to be revisited later in some further work. In the meantime this work should fix the issue as observed by iOS users and ensure the tool is useable by them.

### DIFF
--- a/app/assets/javascripts/components/MoneyNavigatorResults.js
+++ b/app/assets/javascripts/components/MoneyNavigatorResults.js
@@ -30,6 +30,7 @@ define(['jquery', 'DoughBaseComponent', 'utilities'], function($, DoughBaseCompo
     this._updateDOM(); 
     this._setUpEvents(); 
     this._initialisedSuccess(initialised);
+    this.bodyHeight = document.body.clientHeight; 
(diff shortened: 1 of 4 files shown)


akenzua (approved):
Couldn't reproduce the issue (even with browserstack) but the PR did what it set out to do. So I am happy to approve.

PR #2352, opened 2021-07-19
Bump `universal_credit` for embedded version

benlovell:
This moves the money manager dependency on to the latest, production
version themed for Money Helper. In doing so, the authentication links
are duplicated due to them having been moved into the embed content
itself as the expectation is that this tool will now only ever be
presented in its syndicated guise.

I've modified the CSS selection for the `auth` section to just find the
first it comes across, this works around the ambiguous match that
results otherwise.

https://maps.tpondemand.com/entity/12681-make-money-manager-production-ready


jordanlg27:
https://maps.tpondemand.com/entity/12681

PR #2359, opened 2021-07-29
TP-12609 bump WPCC for UAT/production

benlovell:
Includes 12642, 12597, 12632 and 12609.


jordanlg27:
https://maps.tpondemand.com/entity/12609

PR #2361, opened 2021-08-11
Bump gems to merged/production versions

benlovell:
There are outstanding issues with dough compatibility, especially around
the older tools (that are now likely EOL'd) that will be addressed.

https://maps.tpondemand.com/entity/12817-rebasing-front-end-app-with-master


jordanlg27:
https://maps.tpondemand.com/entity/12817

PR #2364, opened 2021-08-13
Fix karma build failures

benlovell:
We've had a weird build failure due to karma specs not being run. This change bumps karma and introduces some more verbosity to the logs to help analyse the failures.


jordanlg27:
https://maps.tpondemand.com/entity/12857

PR #2380, opened 2022-01-04
Set Referrer-Policy to no-referrer-when-downgrade

AndrewVos:
This allows us to see a full URL when looking at HTTP_REFERER
inside the iframe.


AndrewVos:
Closing because this header needs to be added on the AEM side.

PR #2390, opened 2022-01-24
Bump car cost tool gem to remove images

benlovell:
Removes the broken CAP images for vehicles.


benlovell:
Closing in favour of #2391.

